PR #26, opened 2016-04-21
Fixing an int overflow bug

ScottMansfield:
If a client did more than 2^31 requests to a single cache, they would
overflow and get an index out of bounds error during client selection.

This uses the long counter we already had and takes the absolute value
(just in case someone manages to do 9 quintillion requests) and then
mods before truncating to ensure we use the long value and not a
falsely truncated int value.

CC @vuzilla @senugula @smadappa

### DIFF
--- a/evcache-client/src/main/java/com/netflix/evcache/pool/EVCacheClientPool.java
+++ b/evcache-client/src/main/java/com/netflix/evcache/pool/EVCacheClientPool.java
@@ -212,10 +212,10 @@ public class EVCacheClientPool implements Runnable, EVCacheClientPoolMBean {
             return clients.get(0); // Frequently used scenario
         }
 
+        final long currentVal = numberOfModOps.incrementAndGet();
         // Get absolute value of current val to ensure correctness even at 9 quintillion+ requests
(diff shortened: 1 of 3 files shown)


evcache-client/src/main/java/com/netflix/evcache/pool/EVCacheClientPool.java
```diff
@@ -211,8 +211,11 @@ private EVCacheClient selectClient(List<EVCacheClient> clients) {
         if (clients.size() == 1) {
             return clients.get(0); // Frequently used scenario
         }
-        final long currentVal = numberOfModOps.incrementAndGet();
-        final int index = (int) currentVal % clients.size();
+
+        // Get absolute value of current val to ensure correctness even at 9 quintillion+ requests
+        final long currentVal = Math.abs(numberOfModOps.incrementAndGet());
```
vuzilla: can you move the Math.abs to the next line for index.

ScottMansfield:
Checks aren't successful because the Travis setup was done after this branched off of main.

PR #28, opened 2016-07-19
use archaius instead of system properties for SimpleNodeListProvider

trigan-d:
Hi.
We want to use SimpleNodeListProvider and configure the node addresses via archaius. But faced that both provider type and node addresses are read from system properties instead. 
Was it done intentionally? Do you have some reason for that?
If not, then please consider accepting my PR and releasing a new patch (or minor) version.
Thanks in advance!


smadappa:
This was an oversight. Thanks for fixing it.

trigan-d:
Cool. Could you make a release please?

trigan-d:
Cool! Could you make a release please?

PR #70, opened 2020-01-02
Add Zipkin tracing tags for EVCache calls

maulikpandey:
**Context**
We want to add observability to EVCache client calls by creating Zipkin spans and adding relevant tracing tags.
 
**Description**
This PR adds a `evcache-zipkin-tracing` subproject to the EVCache repo that contains a EVCache call listener which creates [Zipkin](https://github.com/openzipkin/zipkin) spans using the [Brave](https://github.com/openzipkin/brave) client.


maulikpandey:
@smadappa Let me know if the changes look good.

evcache-zipkin-tracing/src/main/java/com/netflix/evcache/EVCacheTracingEventListener.java
```diff
@@ -0,0 +1,161 @@
...
+       * attempt to write to its cache instance.
+       */
+      String serverGroup;
+      StringJoiner serverGroups = new StringJoiner(",", "[", "]");
+      for (EVCacheClient client : e.getClients()) {
+        serverGroup = client.getServerGroupName();
+        if (StringUtils.isNotBlank(serverGroup)) {
+          serverGroups.add("\"" + serverGroup + "\"");
+        }
+      }
```
narayaruna: This could apply to any tag value that will be of type list. May be this can be moved to the shared library. Also the `safeTag` method. More than code use, this will bring consistency in how the data is encoded on to the tags.

narayaruna:
Looks good to me. Just added one comment. Another thing possibly is to squash and merge so that it is easy to review what went in to one commit.

evcache-zipkin-tracing/src/main/java/com/netflix/evcache/EVCacheTracingEventListener.java
```diff
@@ -0,0 +1,161 @@
...
+    try {
+      if (t != null) {
+        this.safeTag(clientSpan, EVCacheTracingTags.ERROR, t.toString());
+      }
+
+      String status = e.getStatus();
+      this.safeTag(clientSpan, EVCacheTracingTags.STATUS, status);
+
+      long latency = e.getDurationInMillis();
+      clientSpan.tag(EVCacheTracingTags.LATENCY, String.valueOf(latency));
```
kerumai: Is this latency any different from the implicit span duration? ie. is this duplicating that?

evcache-zipkin-tracing/src/main/java/com/netflix/evcache/EVCacheTracingTags.java
```diff
@@ -0,0 +1,15 @@
+package com.netflix.evcache;
+
+public class EVCacheTracingTags {
```
kerumai: Can we re-use tag names from `EVCacheMetricsFactory` and the IPC metrics where applicable here?

evcache-zipkin-tracing/src/main/java/com/netflix/evcache/EVCacheTracingEventListener.java
maulikpandey: EVCache client does async operations for quorum. Latency from EVCacheEvent will be appropriate here as underlying implementation of EVCache threads may change in future.

evcache-zipkin-tracing/src/main/java/com/netflix/evcache/EVCacheTracingTags.java
maulikpandey: Tag names in `EVCacheMetricsFactory` follow the camel case. All our tags are lower case strings with underscores.

evcache-zipkin-tracing/src/main/java/com/netflix/evcache/EVCacheTracingEventListener.java
kerumai: Is this latency tag duplicating the _duration_ attribute that Brave adds to all spans though?

evcache-zipkin-tracing/src/main/java/com/netflix/evcache/EVCacheTracingTags.java
kerumai: I was thinking of these:
* `evcache.call` vs `evc.call`
* `evcache.app_name` vs `ipc.server.app`
* `evcache.status` vs `ipc.status`
and possibly some of the others.

ie. prefer using existing tag names where they exist, rather than minting new ones.

PR #71, opened 2020-01-07
Fixing tag override problem for multiple cached keys.

maulikpandey:
The EVCacheEvent::getEVCacheKeys() method returns multiple keys in case of bulk operation. It causes the tags to be overwritten by the last iterated key.

This PR fixes the the overwrite issue. Additionally, we are adding an optimization that records tags if tracing is enabled by checking `clientSpan.isNoop()`.


maulikpandey:
cc - @asibross @smadappa @narayaruna

PR #72, opened 2020-01-31
Fix timestamp units from milliseconds to microseconds.

maulikpandey:
**Bug**
`timestamp` field in EVCache traces are coming in with milliseconds as the unit of time. Zipkin specification requires all timestamps to be of microsecond time unit. The `duration` field in the traces are  invalid as a side effect of calculating `endTimeInMicroseconds - startTimeInMilliseconds`.

**Fix**
Use the default Zipkin startTime when starting the client span in `EVCacheEventListener`.  Additionally, we are converting the underlying latency of EVCache call to microseconds as well for adhering to Zipkin specification.


maulikpandey:
cc - @smadappa @narayaruna @kerumai

senugula:
@maulikpandey Looks good.

PR #83, opened 2020-07-13
Consolidating hashing to Impl, and removing hashing from EVCacheClient

sumanth-pasupuleti:
EVCacheKey class handles the logic of overlaying hashing of app level and EVCacheClient level, making it possible for EVCacheImpl to take the hashing decision at both the levels, and thereby removing hashing logic from EVCacheClient

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
@@ -27,7 +27,6 @@ import org.slf4j.LoggerFactory;
 
 import com.netflix.archaius.api.Property;
 import com.netflix.archaius.api.PropertyRepository;
-import com.netflix.evcache.EVCache.Call;
 import com.netflix.evcache.EVCacheInMemoryCache.DataNotFoundException;
 import com.netflix.evcache.EVCacheLatch.Policy;
 import com.netflix.evcache.event.EVCacheEvent;
@@ -90,6 +89,8 @@ final public class EVCacheImpl implements EVCache, EVCacheImplMBean {
     private final Property<Boolean> ignoreTouch;
     private final Property<Boolean> hashKey;
     private final Property<String> hashingAlgo;
+    private final Property<Boolean> shouldEncodeHashKey;
+    private final Property<Integer> maxHashingBytes;
     private final EVCacheTranscoder evcacheValueTranscoder;
     private final Property<Integer> maxReadDuration, maxWriteDuration;
 
@@ -147,6 +148,8 @@ final public class EVCacheImpl implements EVCache, EVCacheImplMBean {
 
         this.hashKey = propertyRepository.get(appName + ".hash.key", Boolean.class).orElse(false);
         this.hashingAlgo = propertyRepository.get(appName + ".hash.algo", String.class).orElse("siphash24");
+        this.shouldEncodeHashKey = propertyRepository.get(appName + ".hash.encode", Boolean.class).orElse(true);
+        this.maxHashingBytes = propertyRepository.get(appName + ".hash.max.bytes", Integer.class).orElse(-1);
         this.autoHashKeys = propertyRepository.get(_appName + ".auto.hash.keys", Boolean.class).orElseGet("evcache.auto.hash.keys").orElse(false);
         this.evcacheValueTranscoder = new EVCacheTranscoder();
         evcacheValueTranscoder.setCompressionThreshold(Integer.MAX_VALUE);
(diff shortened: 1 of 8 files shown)


evcache-core/src/main/java/com/netflix/evcache/EVCacheKey.java
```diff
@@ -99,16 +113,6 @@ public boolean equals(Object obj) {
...
-        if (hashKey == null) {
-            if (other.hashKey != null)
-                return false;
-        } else if (!hashKey.equals(other.hashKey))
-            return false;
-        if (hashKeyForDuet == null) {
-            if (other.hashKeyForDuet != null)
-                return false;
-        } else if (!hashKeyForDuet.equals(other.hashKeyForDuet))
-            return false;
```
sumanth-pasupuleti, line 111: @smadappa I haven't added any equal checks on the new hashmaps that contain hashed keys for duet and non-duet, due to the additional cost of such comparison. Consequently, we are now not considering hashkeys for equal check of EVCacheKey. Let me know if you see any concerns with this.

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
```diff
@@ -977,7 +969,30 @@ private EVCacheItemMetaData getEVCacheItemMetaData(EVCacheClient client, EVCache
...
+            String canonicalKey = evcKey.getCanonicalKey(client.isDuetClient());
+            if (hashKey != null) {
+                final EVCacheItem<Object> obj = client.metaGet(hashKey, evcacheValueTranscoder, throwException, hasZF);
+                if (obj.getData() instanceof EVCacheValue) {
+                    final EVCacheValue val = (EVCacheValue) obj.getData();
+                    if (null == val) {
+                        return null;
+                    }
+                    if (!(val.getKey().equals(canonicalKey))) {
+                        incrementFailure(EVCacheMetricsFactory.KEY_HASH_COLLISION, Call.GET.name(), EVCacheMetricsFactory.META_GET_OPERATION);
```
sumanth-pasupuleti, line 982: @smadappa As I pulled this code from EVCacheClient to Impl, I added the operationType, in addition to the earlier existing operation. jfyi.

evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClientUtil.java
```diff
@@ -25,27 +25,26 @@ public EVCacheClientUtil(EVCacheClientPool pool) {
...
      * TODO : once metaget is available we need to get the remaining ttl from an existing entry and use it 
      */
-    public EVCacheLatch add(EVCacheKey evcKey, final CachedData cd, boolean shouldHashKey, Transcoder evcacheValueTranscoder, int timeToLive, Policy policy) throws Exception {
+    public EVCacheLatch add(EVCacheKey evcKey, final CachedData cd, Transcoder evcacheValueTranscoder, int timeToLive, Policy policy) throws Exception {
         if (cd == null) return null; 
         
         final EVCacheClient[] clients = _pool.getEVCacheClientForWrite();
         final EVCacheLatchImpl latch = new EVCacheLatchImpl(policy, clients.length - _pool.getWriteOnlyEVCacheClients().length, _appName);
 
-        CachedData cd1 = null;
```
sumanth-pasupuleti, line 37: @smadappa I had to pull this variable down, to the level of each client, since otherwise, this can have incorrect influence on other clients.
For instance, if the first client in the for loop has hashing enabled, cd1 contains its hashedkey. Now, for the subsequent client, if hashing is not enabled, we will still end up using cd1 of the previous client.

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
smadappa: @sumanth-pasupuleti Should we change this to `Call.META_GET.name()`?

evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClientUtil.java
smadappa: Looks good. The only thing is we might encode twice in some situations for add esp if we have mixed algo or some copies hashed and some not hashed.

smadappa: This might be ok as I don't expect this to be a common use case. This might be pretty rare if it were to even occur. We can address this if it happens.

evcache-core/src/main/java/com/netflix/evcache/util/KeyHasher.java
```diff
@@ -45,40 +65,40 @@
 //        }
 //    }
 
-    public static String getHashedKey(String key, String hashingAlgorithm) {
+    public static String getHashedKey(String key, HashingAlgorithm hashingAlgorithm) {
```
smadappa, line 68: Let us also add an overloaded method like max hashing bytes. 
For ex, I want to use murmur3 but only first 12 bytes instead of 16 bytes.

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
sumanth-pasupuleti: That makes a lot more sense! Changing it to Call.META_GET.name()

evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClientUtil.java
sumanth-pasupuleti: @smadappa I think I see what you mean. We will not be double encoding or double hashing, but we may be repeating the work we already did of calculating the encoded value, for each of the clients. Yeah, let us revisit this if needed, and as a potential improvement, we can explore storing each encoded value (each for non-duet and duet) and reuse if needed.

evcache-core/src/main/java/com/netflix/evcache/util/KeyHasher.java
sumanth-pasupuleti: done

PR #95, opened 2020-10-08
Added support for Ascii85 encoding

smadappa:


evcache-core/src/main/java/com/netflix/evcache/EVCacheKey.java
```diff
@@ -83,36 +89,52 @@ public String getHashKey(boolean isDuet, HashingAlgorithm hashingAlgorithm, Bool
...
 
+        final String key = hashingAlgorithm.toString()+ maxDigestBytes != null ? maxDigestBytes.toString() : "-" + maxHashLength != null ? maxHashLength.toString() : "-" + encoder != null ? encoder : "-";
+        String val = hashedKeysByAlgorithmForDuet.get(key);
+        if(val == null) {
+            val = KeyHasher.getHashedKeyEncoded(getCanonicalKeyForDuet(), hashingAlgorithm, maxDigestBytes, maxHashLength, encoder);
+            hashedKeysByAlgorithmForDuet.put(key , val);
+        }
         // TODO: Once the issue around passing hashedKey in bytes[] is figured, we will start using (nullable) shouldEncodeHashKey, and call KeyHasher.getHashedKeyInBytes() accordingly
-        return hashedKeysByAlgorithmForDuet.computeIfAbsent(hashingAlgorithm, ha -> KeyHasher.getHashedKeyEncoded(getCanonicalKeyForDuet(), ha, maxDigestBytes, maxHashLength));
+        return null;
```
sumanth-pasupuleti, line 137: return val

evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClient.java
```diff
@@ -153,6 +155,7 @@
         this.shouldEncodeHashKey = EVCacheConfig.getInstance().getPropertyRepository().get(this.serverGroup.getName() + ".hash.encode", Boolean.class).orElse(null);
         this.maxDigestBytes = EVCacheConfig.getInstance().getPropertyRepository().get(this.serverGroup.getName() + ".max.digest.bytes", Integer.class).orElse(null);
         this.maxHashLength = EVCacheConfig.getInstance().getPropertyRepository().get(this.serverGroup.getName() + ".max.hash.length", Integer.class).orElse(null);
+        this.encoderBase = EVCacheConfig.getInstance().getPropertyRepository().get(this.serverGroup.getName() + ".hash.encoder", String.class).orElse("base64");
```
sumanth-pasupuleti, line 158: Maybe we should also include an app level encoder in EVCacheImpl, that will then be used when instantiating EVCacheKey. This will be helpful when we want to use something other than base64 for all the ASGs, without need to add an override for each ASG.

PR #97, opened 2020-10-19
Fixed a bug with hashedkey where we were dueting it by default.

smadappa:


evcache-core/src/main/java/com/netflix/evcache/EVCacheKey.java
```diff
@@ -115,9 +125,10 @@ private String getHashKey(HashingAlgorithm hashingAlgorithm, Boolean shouldEncod
         final String key = hashingAlgorithm.toString()+ maxDigestBytes != null ? maxDigestBytes.toString() : "-" + maxHashLength != null ? maxHashLength.toString() : "-" + encoder != null ? encoder : "-";
         String val = hashedKeysByAlgorithm.get(key);
         if(val == null) {
-            val = KeyHasher.getHashedKeyEncoded(getCanonicalKeyForDuet(), hashingAlgorithm, maxDigestBytes, maxHashLength, encoder);
+            val = KeyHasher.getHashedKeyEncoded(canonicalKey, hashingAlgorithm, maxDigestBytes, maxHashLength, encoder);
```
sumanth-pasupuleti, line 128: maybe good to use accessor getCanonicalKey()

PR #104, opened 2021-03-22
EVCACHE-454: allow to ignore eureka status of application. 

arunagrawal84:
Ideally, the app should not serve traffic until it is healthy but this option - ignoreAppEurekaStatus will allow the app to ignore application eureka status while trying to initialize evcache client. This feature `may` be required for applications that are behind load balancers. They may fail to register successfully with eureka but are still expected to serve traffic.

### DIFF
--- a/evcache-client/src/main/java/com/netflix/evcache/pool/eureka/EurekaNodeListProvider.java
+++ b/evcache-client/src/main/java/com/netflix/evcache/pool/eureka/EurekaNodeListProvider.java
@@ -56,7 +56,14 @@ public class EurekaNodeListProvider implements EVCacheNodeList {
      */
     @Override
     public Map<ServerGroup, EVCacheServerGroupConfig> discoverInstances(String _appName) throws IOException {
-        if ((applicationInfoManager.getInfo().getStatus() == InstanceStatus.DOWN)) {
+        final Property<Boolean> ignoreAppEurekaStatus = props.get("evcache.ignoreAppEurekaStatus", Boolean.class).orElse(false);
+
+        if (ignoreAppEurekaStatus.get())
+            log.info("Not going to consider the eureka status of the application, to initialize evcache client.");
+
+        if (!ignoreAppEurekaStatus.get() && (applicationInfoManager.getInfo().getStatus() == InstanceStatus.DOWN)) {
+            log.info("Not initializing evcache client as application eureka status is DOWN. " +
+                    "One can override this behavior by setting evcache.ignoreAppEurekaStatus on application.");
             return Collections.<ServerGroup, EVCacheServerGroupConfig> emptyMap();
         }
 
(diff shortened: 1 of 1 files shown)


evcache-client/src/main/java/com/netflix/evcache/pool/eureka/EurekaNodeListProvider.java
```diff
@@ -56,7 +56,14 @@ public EurekaNodeListProvider(ApplicationInfoManager applicationInfoManager, Eur
...
     public Map<ServerGroup, EVCacheServerGroupConfig> discoverInstances(String _appName) throws IOException {
-        if ((applicationInfoManager.getInfo().getStatus() == InstanceStatus.DOWN)) {
+        final Property<Boolean> ignoreAppEurekaStatus = props.get("evcache.ignoreAppEurekaStatus", Boolean.class).orElse(false);
+
+        if (ignoreAppEurekaStatus.get())
+            log.info("Not going to consider the eureka status of the application, to initialize evcache client.");
+
+        if (!ignoreAppEurekaStatus.get() && (applicationInfoManager.getInfo().getStatus() == InstanceStatus.DOWN)) {
+            log.info("Not initializing evcache client as application eureka status is DOWN. " +
+                    "One can override this behavior by setting evcache.ignoreAppEurekaStatus on application.");
```
sumanth-pasupuleti, line 66: Also maybe include value for the property, like
"...by setting evcache.ignoreAppEurekaStatus property to true, scoped to your application."

evcache-client/src/main/java/com/netflix/evcache/pool/eureka/EurekaNodeListProvider.java
```diff
@@ -56,7 +56,14 @@ public EurekaNodeListProvider(ApplicationInfoManager applicationInfoManager, Eur
      */
     @Override
     public Map<ServerGroup, EVCacheServerGroupConfig> discoverInstances(String _appName) throws IOException {
-        if ((applicationInfoManager.getInfo().getStatus() == InstanceStatus.DOWN)) {
+        final Property<Boolean> ignoreAppEurekaStatus = props.get("evcache.ignoreAppEurekaStatus", Boolean.class).orElse(false);
+
+        if (ignoreAppEurekaStatus.get())
+            log.info("Not going to consider the eureka status of the application, to initialize evcache client.");
```
sumanth-pasupuleti, line 62: i would recommend changing this to debug log, since this isn't a very common case of applications wanting to do useful work, while being down in discovery.

sumanth-pasupuleti (approved):
Have a couple of minor logging statement comments. Otherwise, LGTM.

evcache-client/src/main/java/com/netflix/evcache/pool/eureka/EurekaNodeListProvider.java
arunagrawal84: This log gets printed only when someone has enabled the property. I wanted to ensure we print some INFO log to ensure we "capture" this if we are looking into some issue, so we know about the extra behavior.

PR #107, opened 2021-03-25
Adding a basic Meta arithmetic implementation 

tharanga:
This will get binary incr/decr compatibility for the ascii protocol

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/operation/EVCacheAsciiOperationFactory.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/operation/EVCacheAsciiOperationFactory.java
@@ -7,6 +7,11 @@ import net.spy.memcached.protocol.ascii.MetaDebugOperation;
 import net.spy.memcached.protocol.ascii.MetaDebugOperationImpl;
 import net.spy.memcached.protocol.ascii.MetaGetOperation;
 import net.spy.memcached.protocol.ascii.MetaGetOperationImpl;
+import net.spy.memcached.protocol.ascii.MetaArithmeticOperationImpl;
+import net.spy.memcached.ops.Mutator;
+import net.spy.memcached.ops.MutatorOperation;
+import net.spy.memcached.ops.OperationCallback;
+
 
 public class EVCacheAsciiOperationFactory extends AsciiOperationFactory {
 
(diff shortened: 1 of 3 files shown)


evcache-core/src/main/java/com/netflix/evcache/operation/EVCacheAsciiOperationFactory.java
```diff
@@ -23,5 +28,9 @@ public ExecCmdOperation execCmd(String cmd, ExecCmdOperation.Callback cb) {
         return new ExecCmdOperationImpl(cmd, cb);
       }
 
+    public MutatorOperation mutate(Mutator m, String key, long by, long def,
```
arunagrawal84, line 31: nit: would prefer descriptive names like incrementAmount, defaultValue, expiry, opCallback.

evcache-core/src/main/java/net/spy/memcached/protocol/ascii/MetaArithmeticOperationImpl.java
```diff
@@ -0,0 +1,116 @@
...
+import net.spy.memcached.ops.StatusCode;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.nio.ByteBuffer;
+import java.util.Collection;
+import java.util.Collections;
+
+public class MetaArithmeticOperationImpl extends EVCacheOperationImpl implements
```
sumanth-pasupuleti, line 18: nit: good to include description for public classes

evcache-core/src/main/java/net/spy/memcached/protocol/ascii/MetaArithmeticOperationImpl.java
```diff
@@ -0,0 +1,116 @@
...
+        mutator = m;
+        key = k;
+        amount = amt;
+        def = defaultVal;
+        exp = expiry;
+        readingValue = false;
+    }
+
+    @Override
+    public void handleLine(String line) {
```
sumanth-pasupuleti, line 59: given that core of the logic in the class is protocol implementation, I am not reviewing the logic itself.

sumanth-pasupuleti (approved):

evcache-core/src/main/java/com/netflix/evcache/operation/EVCacheAsciiOperationFactory.java
tharanga: I kept the nomenclature similar to other op implementations, such as MutatorOperationImpl.

PR #113, opened 2021-04-29
Automatically switch logs to DEBUG on client wrong key error

smukil:
New Boolean FP to control log switching:
<appName>.enable.auto.log.switch

NOTE: This is only a temporary commit to help catch a bug. This
will be reverted once we've fixed the underlying bug.

### DIFF
--- a/evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
+++ b/evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
@@ -76,6 +76,7 @@ public class EVCacheMemcachedClient extends MemcachedClient {
     private Property<Long> mutateOperationTimeout;
     private final ConnectionFactory connectionFactory;
     private final Property<Integer> maxReadDuration, maxWriteDuration;
+    private final Property<Boolean> autoLogSwitchOnWrongKey;
 
     public EVCacheMemcachedClient(ConnectionFactory cf, List<InetSocketAddress> addrs,
                                   Property<Integer> readTimeout, EVCacheClient client) throws IOException {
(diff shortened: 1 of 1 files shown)


pkarumanchi9 (approved):

evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
```diff
@@ -86,6 +87,7 @@ public EVCacheMemcachedClient(ConnectionFactory cf, List<InetSocketAddress> addr
         this.appName = client.getAppName();
         this.maxWriteDuration = EVCacheConfig.getInstance().getPropertyRepository().get(appName + ".max.write.duration.metric", Integer.class).orElseGet("evcache.max.write.duration.metric").orElse(50);
         this.maxReadDuration = EVCacheConfig.getInstance().getPropertyRepository().get(appName + ".max.read.duration.metric", Integer.class).orElseGet("evcache.max.read.duration.metric").orElse(20);
+        this.autoLogSwitchOnWrongKey = EVCacheConfig.getInstance().getPropertyRepository().get(appName + ".enable.auto.log.switch", Boolean.class).orElse(false);
```
tharanga, line 90: nit: some text to indicate this is for specific for the wrong key error would be helpful, something like "enable.debug.logs.on.wrongkey"

evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
```diff
@@ -111,6 +113,11 @@ private boolean isWrongKeyReturned(String original_key, String returned_key) {
             log.error("Wrong key returned. Key - " + original_key + " (Host: " + original_host + ") ; Returned Key "
                         + returned_key + " (Host: " + returned_host + ")", new Exception());
             client.reportWrongKeyReturned(original_host);
+
+            // If we are configured to dynamically switch log levels to DEBUG on a wrong key error, do so here.
+            if (autoLogSwitchOnWrongKey.get()) {
+                org.apache.log4j.Logger.getRootLogger().setLevel(org.apache.log4j.Level.toLevel("DEBUG"));
```
tharanga, line 119: Which logger is the root logger in this context? We want to limit debug logs to "log4j.logger.net.spy.memcached" or know the name of the logger we are enabling here in order to disable it.

evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
smukil: Done.

evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
smukil: @tharanga This is a little unclear to me as well. Would something like this work?
org.apache.log4j.Logger.getLogger(EVCacheMemcachedClient.class).setLevel(org.apache.log4j.Level.toLevel("DEBUG"));

But this might only make logs DEBUG for the EVCacheMemcachedClient class.

smukil: I changed it to ```System.setProperty("log4j.logger.net.spy", "DEBUG");``` as per @arunagrawal84's suggestion.

tharanga (approved):

PR #122, opened 2022-04-26
Adding a config/FP to ignore the set alias. 

tharanga:
This helps EVCache router clusters to use the correct backend regardless of the set alias.


evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClientPoolManager.java
```diff
@@ -251,7 +251,14 @@ public EVCacheExecutor getEVCacheExecutor() {
 
     private String getAppName(String _app) {
         _app = _app.toUpperCase();
-        final String app = EVCacheConfig.getInstance().getPropertyRepository().get("EVCacheClientPoolManager." + _app + ".alias", String.class).orElse(_app).get().toUpperCase();
+        Boolean ignoreAlias = EVCacheConfig.getInstance().getPropertyRepository()
+                .get("EVCacheClientPoolManager." + _app + ".ignoreAlias", Boolean.class)
+                .orElseGet("EVCacheClientPoolManager.ignoreAlias")
+                .orElse(false).get();
+        final String app = ignoreAlias ?  _app.toUpperCase() :
```
pkarumanchi9, line 258: Just _app is good as line 253 is already converting to upper case.

And line 263 is also returning the uppercase.

pkarumanchi9 (approved):

PR #123, opened 2022-05-11
use FP to choose if optimize broadcast

zhitaozhu:
Thanks for the fix.

Context:
we are mutating a strings collection while iterating over the collection. This happens specifically during the addOperation as part of OptimizedGetImpl.java


pkarumanchi9:
After digging in further, we noticed that this optimized path can be toggled via a variable. But that is hard-coded to true all the time. We thought of rolling in a fix where we can make it a configurable parameter via fp. The idea here is to disable the optimized get path and let the app team pick this version of evcache-client. Since they are able to repeatedly see this behavior, we can narrow the problem down to this codepath and go from there.

tharanga:
How are we planning to use this FP?

PR #128, opened 2022-07-12
router client changes for read retries

sriram-rangarajan:
router client changes for read retries


pkarumanchi9 (approved):
Looks good. Ship It!

PR #129, opened 2022-07-12
router write changes. No writing to all replicas

sriram-rangarajan:
router write changes. No writing to all replicas

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
@@ -96,6 +96,7 @@ public class EVCacheImpl implements EVCache, EVCacheImplMBean {
     private final EVCacheTranscoder evcacheValueTranscoder;
     private final Property<Integer> maxReadDuration, maxWriteDuration;
     private final Property<Boolean> clientReadRetry;
+    private final Property<Boolean> clientWriteToAllReplicas;
 
     protected final EVCacheClientPoolManager _poolManager;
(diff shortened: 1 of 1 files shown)


evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
```diff
@@ -2210,12 +2212,27 @@ public int getDefaultTTL() {
         return _timeToLive;
     }
 
+    private EVCacheClient[] getClientsForWrite() {
+        EVCacheClient[] clients;
+        if (clientWriteToAllReplicas.get()) {
+            clients = _pool.getEVCacheClientForWrite();
+        } else {
+            EVCacheClient client = _pool.getEVCacheClientForRead();
+            if (client == null) {
```
pkarumanchi9, line 2221: Can you please add a comment on why client can be null?

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
```diff
@@ -2210,12 +2212,27 @@ public int getDefaultTTL() {
         return _timeToLive;
     }
 
+    private EVCacheClient[] getClientsForWrite() {
+        EVCacheClient[] clients;
+        if (clientWriteToAllReplicas.get()) {
```
pkarumanchi9, line 2217: If router instance is down or the tcp connection is severed, we need to handle the retries. For now, lets push these changes and release a candidate build.

pkarumanchi9 (approved):
Let's release a candidate build with these changes.

PR #130, opened 2022-09-27
disable zone affinity and fix some legacy code around rend

pkarumanchi9:
Fixes -

1. introduces an fp to disable zone affinity
2. adds metric to track the amount of time spent in eureka bootstrap (debug issue with refreshPools)
3. removes legacy code around rend


sriram-rangarajan (approved):
LGTM

PR #132, opened 2022-12-14
Sort Client (Read/Write ASG followed by Write only) for add API

sriram-rangarajan:
Context: 
For ADD api we skip adding keys to the cluster if the key is present in the first client. Now if a new cluster is warmed with fewer keys, the ADDS might not go through as it doesn't respect read/write mode and this might result in drop in hit rate. 

Fix:
Sorting the client (read/write ASG first then followed by write-only ASG). So now the ADD API with check if the key is present in ASG  which is actually serving the traffic rather than write-only ASG. 

Please let me know if you have any questions. Please review and let me know your comments. Thanks!

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
@@ -20,6 +20,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import javax.management.MBeanServer;
 import javax.management.ObjectName;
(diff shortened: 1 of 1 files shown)


arunagrawal84 (approved):
LGTM

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
```diff
@@ -3259,7 +3260,21 @@ public <T> boolean add(String key, T value, Transcoder<T> tc, int timeToLive) th
     @Override
     public <T> EVCacheLatch add(String key, T value, Transcoder<T> tc, int timeToLive, Policy policy) throws EVCacheException {
         EVCacheClient[] clients = _pool.getEVCacheClientForWrite();
-        return this.add(key, value, tc, timeToLive, policy, clients, clients.length - _pool.getWriteOnlyEVCacheClients().length);
+        EVCacheClient[] writeOnlyClients = _pool.getWriteOnlyEVCacheClients();
+        // In case of adds , we skip adds to the pool if value is already present in the 1st client
```
akashdeepgoel, line 3264: Any concerns with enabling the same behavior as SET calls? Fan out writes to all ASGs that can take writes.

PR #139, opened 2023-06-28
Add percentile timer metric for gzip compression ratio

samuelfu:
EVCacheSerializingTranscoder is an exact copy of SerializingTranscoder, a transcoder class in Spymemcached. This pull request allows EVCacheSerializingTranscoder to add functionalities in the future, including logging, and custom encoding/decoding.

Compression ratio is defined as 
```
long compression_ratio = Math.round((double) compressed.length / b.length * 100);
```

The ratio is converted into long as a percentage since timer only takes long or Duration as argument.

PercentileTimer is used to log the compression ratios for all writes > 120 bytes.

The unit test compares the cached data encoded data with SerializingTranscoder's cached data encoded object and passes if they are equal.

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
@@ -22,13 +22,24 @@
 
 package com.netflix.evcache;
 
+import com.netflix.evcache.metrics.EVCacheMetricsFactory;
+import com.netflix.evcache.pool.ServerGroup;
+import com.netflix.spectator.api.BasicTag;
+import com.netflix.spectator.api.Tag;
+import com.netflix.spectator.api.Timer;
 import net.spy.memcached.CachedData;
 import net.spy.memcached.transcoders.BaseSerializingTranscoder;
 import net.spy.memcached.transcoders.Transcoder;
 import net.spy.memcached.transcoders.TranscoderUtils;
 import net.spy.memcached.util.StringUtils;
 
+import java.time.Duration;
+import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


evcache-core/src/main/java/com/netflix/evcache/metrics/EVCacheMetricsFactory.java
```diff
@@ -251,6 +251,7 @@ public String getStatusCode(StatusCode sc) {
      */
     public static final String OVERALL_CALL                         = "evcache.client.call";
     public static final String OVERALL_KEYS_SIZE                    = "evcache.client.call.keys.size";
+    public static final String COMPRESSION_RATIO                    = "evcache.client.compression_ratio";
```
pkarumanchi9, line 254: change _ to .

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    private void updateTimerWithCompressionRatio(long ratio_percentage) {
+        if(timer == null) {
+            final List<Tag> tagList = new ArrayList<Tag>(1);
+
+            timer = EVCacheMetricsFactory.getInstance().getPercentileTimer(EVCacheMetricsFactory.COMPRESSION_RATIO, tagList, Duration.ofMillis(100));
+        };
+
+//        if(serverGroup != null) {
```
pkarumanchi9, line 208: Let's add a tag to include the compression type and we don't need the servergroup/zone one's as they get published automatically.

pkarumanchi9 (approved):

PR #141, opened 2023-07-18
Fix compilation issues

rpalcolea:


pkarumanchi9:
Thanks for sending the PR @rpalcolea 
cc: @sriram-rangarajan

PR #144, opened 2024-01-11
Change from debug/info to trace for compression

samuelfu:
Info logs per compression were affecting performance, reducing logs to trace level.


arunagrawal-84:
LGTM

PR #145, opened 2024-05-30
Expose a setter in EVCache.Builder to explicitly set the pool manager

asibross:
Currently the pool manager is implicitly set in the EVCache.Builder instance using DI injection.
This effectively means that the EVCache.Builder must be an instance managed by a DI container.

Exposing this setter enables creating EVCache.Builder instances that is not managed by a container.


pkarumanchi9 (approved):

sriram-rangarajan (approved):
LGTM

PR #148, opened 2024-09-29
Fix build issues related to missing grgit-core artifact

kilink:
Force grgit-core to 4.1.1 to resolve build issues. This is a temporary workaround, the ideal solution is to move the project to the latest gradle and netflixoss plugin versions, but that is a larget undertaking.


pkarumanchi9 (approved):

srrangarajan:
LGTM

PR #151, opened 2024-11-14
perf: skip future cancellation when not needed

jasonk000:
It implicitly throws an expensive exception. Some codepaths throw away the errorFuture, so there is no need to issue the cancel/throw the exception.

This is approximately half of the CPU overhead on the caller thread for the `getAsync` call.

image


pkarumanchi9 (approved):
LGTM!

PR #152, opened 2024-11-18
perf: perform the getPrimary lookup only once during the bulk path

jasonk000:
We can avoid doing the (potentially) expensive getPrimary call on the bulk get path with a little refactor. This retains the same functionality but pushes the validation logic and error emit into a callback, which allows metrics to work as-is.

To do this we introduce a predicate callback that getBulk lower layer can use to validate. In this predicate - which the upper layer provides - we have the appropriate logging and metrics context to make the decision.

In this particular application, the duplicated getPrimary calls are taking the vast majority of the time (I'll have a separate PR for this).

image


jasonk000:
closed in favor of #159

PR #153, opened 2024-11-18
Enhance log for key size check

shy-1234:
The original error message lacked the information for end users to take action. Also added a log so that the information can still be accessed even when the exception is swallowed.

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
@@ -215,7 +215,9 @@ public class EVCacheImpl implements EVCache, EVCacheImplMBean {
         }
 
         if (canonicalKey.length() > this.maxKeyLength.get() && !hashKey.get() && !autoHashKeys.get()) {
-            throw new IllegalArgumentException("Key is too long (maxlen = " + this.maxKeyLength.get() + ')');
+            final String errMsg = "CanonicalKey ``" + canonicalKey + "`` is too long (maxLen = " + this.maxKeyLength.get() + ", keyLen = " + key.length() + ", canonicalKeyLen = " + canonicalKey.length() + ')';
+            log.error(errMsg);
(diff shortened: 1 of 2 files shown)


evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
```diff
@@ -215,7 +215,9 @@ EVCacheKey getEVCacheKey(final String key) {
         }
 
         if (canonicalKey.length() > this.maxKeyLength.get() && !hashKey.get() && !autoHashKeys.get()) {
-            throw new IllegalArgumentException("Key is too long (maxlen = " + this.maxKeyLength.get() + ')');
+            final String errMsg = "CanonicalKey ``" + canonicalKey + "`` is too long (maxLen = " + this.maxKeyLength.get() + ", keyLen = " + key.length() + ", canonicalKeyLen = " + canonicalKey.length() + ')';
+            log.error(errMsg);
```
srrangarajan, line 219: I would recommend to use warn instead of error

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
```diff
@@ -215,7 +215,9 @@ EVCacheKey getEVCacheKey(final String key) {
         }
 
         if (canonicalKey.length() > this.maxKeyLength.get() && !hashKey.get() && !autoHashKeys.get()) {
-            throw new IllegalArgumentException("Key is too long (maxlen = " + this.maxKeyLength.get() + ')');
+            final String errMsg = String.format("CanonicalKey ``%s`` is too long (maxLen = %d, keyLen = %d, canonicalKeyLen = %d)", canonicalKey, this.maxKeyLength.get(), key.length(), canonicalKey.length());
```
brewsterw, line 218: is it possible to add the evcache_app that is causing this? the one that is being written to or read from

PR #154, opened 2024-11-21
Provide pluggable hashing and lookup strategy for NodeLocator

jasonk000:
A significant part of the `getBulk` with large keylists is spent performing NodeLocator lookups (`getPrimary`) to identify which node to send traffic to. The current Ketama configuration is based on MD5 which is no longer best-in-class, and uses a Java TreeMap which is not the most efficient approach for large arrays.

This PR makes the components of the NodeLocator pluggable with different hash/ring functions and different strategies for storage of the hash-ring-to-node lookup function. It then provides a number of alternative implementations. It does NOT change any of the existing implementation / configuration, which would need to come in a follow-up.

Implementation configuration choices range from 30% faster to ~10x faster.

image

### DIFF
--- a/evcache-core/build.gradle
+++ b/evcache-core/build.gradle
@@ -1,5 +1,10 @@
+plugins {
+    id "me.champeau.jmh" version "0.6.6"
+}
+
 apply plugin: 'java'
 apply plugin: 'eclipse'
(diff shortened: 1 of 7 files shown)


jasonk000:
closed in favor of #160

PR #155, opened 2024-12-06
Introduce retries when pinging servers.

pkarumanchi9:
Validate the version response as opposed to logging and moving on.


evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClientPool.java
```diff
@@ -1173,9 +1173,38 @@ public void pingServers() {
...
+                            if (log.isDebugEnabled()) {
+                                for (Entry<SocketAddress, String> vEntry : versions.entrySet()) {
+                                    log.debug("Host : {} Version : {}", vEntry.getKey(), vEntry.getValue());
+                                }
+                            }
+                            break;
+                        }
+
+                        if (i < maxRetries - 1) {
+                            Thread.sleep(retryDelayMs);
```
jasonk000, line 1204: This would block for 10 seconds _per node_ during startup, which could be a very long time and would prevent application startup. I think we should aim to do this asynchronously or some other way.

pkarumanchi9: true, it can block for o(seconds). I've few more thoughts. I will share them shortly.

PR #157, opened 2025-01-27
perf: Avoid charset lookup in key validation

jasonk000:
The upstream implementation needs to do a "UTF-8" string lookup to Charset for every key validation. We can avoid this completely in our hot path by using a reference to StandardCharsets.


pkarumanchi9 (approved):
Looks good!

PR #159, opened 2025-01-29
perf: perform the getPrimary lookup only once during getBulk

jasonk000:
We can avoid doing the (potentially) expensive getPrimary call twice on the bulk get path with a little refactor. This retains the same functionality but pushes the validation logic and error emit into a callback, which allows metrics to work as-is.

before

image

after

image

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClient.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClient.java
@@ -17,6 +17,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.*;
+import java.util.function.BiPredicate;
 import java.util.zip.CRC32;
 import java.util.zip.Checksum;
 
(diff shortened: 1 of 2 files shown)


evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClient.java
```diff
@@ -187,39 +188,36 @@ public Integer getMaxHashLength() {
...
-        if (evcacheMemcachedClient.getNodeLocator() == null) return canonicalKeys;
-        final Collection<String> retKeys = new ArrayList<>(canonicalKeys.size());
-        for (String key : canonicalKeys) {
-            final MemcachedNode node = evcacheMemcachedClient.getNodeLocator().getPrimary(key);
-            if (node instanceof EVCacheNode) {
-                final EVCacheNode evcNode = (EVCacheNode) node;
-                if (!evcNode.isAvailable(call)) {
-                    continue;
-                }
+    private boolean validateReadQueueSize(MemcachedNode node, String key, EVCache.Call call) {
```
pkarumanchi9, line 191: nit:key can be removed? and from all the callers too.

Passing key here doesn't seem to serve any purpose. We want to iterate over all the keys and that's achieved when calling this.

pkarumanchi9 (approved):
LGTM!

evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClient.java
jasonk000: Agreed, fixed in c0c33cf !

PR #161, opened 2025-01-30
perf: track bulk get operation completions explicitly

jasonk000:
This commit makes it so that the operation callbacks are used to collect all the locked state that we might need to access, which makes the code after latch-release run without contending memcached to complete.

Before this commit, callbacks on the bulk get operation accumulate the state and when all operations have completed, we release the latch and then the calller calculates the final state. Unfortunately looking at the state, the caller must take operation locks, which are potentially still hold by memcache driver since it has only momentarily before released the latch.

The Memcache callbacks are made whilst holding the lock on the Operation objects, which means they are a great time to get locked state such as cancellation state. This patch makes the change to collect state during callbacks. The callback future interaction is a little clunky, but this avoids a major rewrite of the implementations.

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/operation/EVCacheBulkGetFuture.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/operation/EVCacheBulkGetFuture.java
@@ -13,10 +13,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicReferenceArray;
 
 import com.netflix.evcache.EVCacheGetOperationListener;
+import com.netflix.evcache.util.Pair;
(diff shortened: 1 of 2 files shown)


evcache-core/src/main/java/com/netflix/evcache/operation/EVCacheBulkGetFuture.java
```diff
@@ -51,6 +54,7 @@ public class EVCacheBulkGetFuture<T> extends BulkGetFuture<T> {
     private final CountDownLatch latch;
     private final long start;
     private final EVCacheClient client;
+    private AtomicReferenceArray<SingleOperationState> operationStates;
```
pkarumanchi9, line 57: Nice!

pkarumanchi9 (approved):

PR #162, opened 2025-01-30
perf: allow concurrent decompress away from network loop

jasonk000:
We want to ensure the transcode passes everything that is expected to be run asynchronously over to the decode loop. In general, memcached calls us back with gotData, then receivedStatus, then complete. We use gotData to launch the work onto the transcode threadpool and return control to memcached, which would immediately then call receivedStatus.

Previously, receivedStatus and complete were set up to interact and set a value on the underlying future but only by synchronously blocking for the transcode future. Given this callback is happening nearly immediately after the gotData callback, we were firing the transcode and nearly always performing a blocking get, which triggers a synchronous decompression on the network IO loop. This is of course very detrimental to evcache driver performance, since the driver cannot even accept new requests to issue to them to memcached backends, and must wait until decompression completes.

In this fix, we rearrange things a little to ensure that if the async decode is requested, that we push the completion status updates to happen only after the async decode completes. This is a little ugly because of the current arrangement of the memcached decoder. A future change might be to overhaul this integration and pull it out of the memcached transcode framework and use something a bit more friendly.

### DIFF
--- a/evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
+++ b/evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
@@ -13,14 +13,16 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiPredicate;
+import java.util.function.Consumer;
 
-import com.netflix.evcache.pool.EVCacheClientUtil;
+import com.netflix.archaius.api.PropertyRepository;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -35,6 +37,7 @@ import com.netflix.evcache.operation.EVCacheItemMetaData;
 import com.netflix.evcache.operation.EVCacheLatchImpl;
 import com.netflix.evcache.operation.EVCacheOperationFuture;
 import com.netflix.evcache.pool.EVCacheClient;
+import com.netflix.evcache.pool.EVCacheClientUtil;
 import com.netflix.evcache.util.EVCacheConfig;
 import com.netflix.spectator.api.BasicTag;
 import com.netflix.spectator.api.DistributionSummary;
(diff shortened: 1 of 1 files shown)


jasonk000:
This change is tricky, actually, since it slightly changes timing it makes the system more scalable without tuning but a little less efficient. It needs some consideration. Maybe an alternative is to always do it in-thread and encourage users to tune poolSize much higher.

evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
```diff
@@ -124,62 +127,85 @@ private boolean isWrongKeyReturned(String original_key, String returned_key) {
...
             public void gotData(String k, int flags, byte[] data) {
+                decodingData = true;
+                if (isWrongKeyReturned(key, k)) {
+                    latch.countDown();
+                    return;
+                }
 
-                if (isWrongKeyReturned(key, k)) return;
+                boolean shouldLog = log.isDebugEnabled() && client.getPool().getEVCacheClientPoolManager().shouldLog(appName);
+                boolean alwaysSync = false;
```
pkarumanchi9, line 153: should be exposed via a fp to go via async path?

evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
```diff
@@ -124,62 +127,85 @@ private boolean isWrongKeyReturned(String original_key, String returned_key) {
...
-                        } else {
-                            final Transcoder<T> t = (Transcoder<T>) getTranscoder();
-                            val = tcService.decode(t, new CachedData(flags, data, t.getMaxSize()));
-                        }
+
+                    if (shouldLog) log.debug("Key : {}; val size : {}", key, data.length);
+                    dataSizeDS.record(data.length);
+                    Transcoder<T> transcoder = (tc == null) ? (Transcoder<T>) getTranscoder() : tc;
+                    if (tcService == null) {
+                        log.error("tcService is null, will not be able to decode");
```
pkarumanchi9, line 162: should we do latch.countDown() here?

evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
jasonk000: fixed in 2b60f17

evcache-core/src/main/java/net/spy/memcached/EVCacheMemcachedClient.java
jasonk000: fixed in 2b60f17, I restructured it too. Hopefully reduce chance of bugs like this.

pkarumanchi9 (approved):
Could you please rebase and merge.

PR #166, opened 2025-06-16
client shutdown without wait time when bean is destroyed

shy-1234:
A couple of clients hung up can cause this to exceed the shutdown limit.


srrangarajan (approved):
LGTM

PR #172, opened 2025-08-15
fix: handle null states in EVCacheBulkGetFuture to prevent NPE

pkarumanchi9:
Fixes NullPointerException in bulk get operations when some operations
  never signal completion through callbacks. The issue was introduced in
  commit e4a2a9bd which optimized operation state checking by collecting
  state during callbacks, but didn't handle cases where callbacks don't fire.

  This fix adds a fallback mechanism that preserves the original behavior:
  - Use pre-collected state when available (maintains performance optimization)
  - Fall back to direct operation state checking when state is null
  - Ensures all operations are processed, matching pre-e4a2a9bd behavior

  The root cause was that operationStates[i] could remain null if
  signalSingleOpComplete() was never called for operations that timeout,
  fail, or are cancelled before their completion callbacks fire.

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/operation/EVCacheBulkGetFuture.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/operation/EVCacheBulkGetFuture.java
@@ -120,13 +120,31 @@ public class EVCacheBulkGetFuture<T> extends BulkGetFuture<T> {
             }
 
             boolean hadTimedoutOp = false;
+            Operation[] opsArray = ops.toArray(new Operation[0]);
             for (int i = 0; i < operationStates.length(); i++) {
                 SingleOperationState state = operationStates.get(i);
-                if (!state.completed && !allCompleted) {
-                    MemcachedConnection.opTimedOut(state.op);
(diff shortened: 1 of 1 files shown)


evcache-core/src/main/java/com/netflix/evcache/operation/EVCacheBulkGetFuture.java
```diff
@@ -120,25 +120,55 @@ public Map<String, T> getSome(long to, TimeUnit unit, boolean throwException, bo
...
             for (int i = 0; i < operationStates.length(); i++) {
                 SingleOperationState state = operationStates.get(i);
-                if (!state.completed && !allCompleted) {
-                    MemcachedConnection.opTimedOut(state.op);
-                    hadTimedoutOp = true;
+                Operation op = opsArray[i];
+                
+                if (state == null) {
+                    // Operation never signaled completion, fall back to direct checking
+                    if (op.getState() != OperationState.COMPLETE) {
```
shy-1234, line 130: Isn't this the same as (the reason why) state == null? Do we ever go into line 138?

PR #189, opened 2026-03-24
Separate inactive node and write queue full metric

shy-1234:
Just to follow up on this [PR comment](https://github.com/Netflix/EVCache/pull/186/changes#r2914011851). Not sure how useful it is, but I think it's good to have this separation.


AustinWheel (approved):
Agree, the distinction is better to show that the node could just have some write backpressure

PR #192, opened 2026-05-14
Add a metric that exposes IO thread utilization

jasonk000:
This allows to spot issues where evcache is overloaded and supports some basic prediction of whether an instance might become overloaded with a significant traffic increase or whether the thread pool could be tuned.

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheLoopProbe.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheLoopProbe.java
@@ -1,0 +1,133 @@
+package com.netflix.evcache.pool;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadMXBean;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Publishes event-loop CPU utilization from the loop thread itself.
+ *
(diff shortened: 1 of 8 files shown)


evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheLoopProbe.java
```diff
@@ -0,0 +1,133 @@
...
+
+        final long now = System.nanoTime();
+        if (nextPublishNs != 0L && now - nextPublishNs < 0L) return;
+        nextPublishNs = now + PUBLISH_INTERVAL_NS;
+
+        final long cpuNs = THREAD_MX_BEAN.getCurrentThreadCpuTime();
+        if (cpuNs < 0L) {
+            if (!negativeCpuTimeLogged) {
+                negativeCpuTimeLogged = true;
+                log.warn("Thread CPU time returned a negative value; skipping EVCache loop CPU utilization publish");
```
akhaku: I'm a little confused as to what we're measuring in this PR - is this different from looking the values reported for CPU time on a thread dump?

jasonk000: Yes, that's exactly it. Exposing this as a metric allows us to look at it easily and assess this as a potential tuning/hotspot if we move a client to a larger instance type (e.g. will 2x the traffic lead to congestion at evcache client).

jasonk000:
I renamed the metric here to be a bit more intuitive. Ran it as a canary on BQS and looks good. Metrics come through. 👍.

PR #193, opened 2026-05-14
Add a metric to capture write enqueue latency

jasonk000:
This captures from when the caller tries to send a request until the request is finally sent on the wire, primarily so we can see when the enqueue time is increasing (a signal for pressure on the IO loop).

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClient.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClient.java
@@ -23,6 +23,7 @@ import com.netflix.spectator.api.Counter;
 import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spectator.api.Tag;
+import com.netflix.spectator.api.Timer;
 import com.netflix.spectator.api.patterns.PolledMeter;
 import java.io.BufferedInputStream;
 import java.io.IOException;
@@ -32,6 +33,7 @@ import java.net.Socket;
 import java.net.SocketAddress;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.util.AbstractMap.SimpleEntry;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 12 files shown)


evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClient.java
```diff
@@ -1695,6 +1701,27 @@ public Counter getOperationCounter() {
...
+     * break an operation completion callback.
+     */
+    public void recordLoopEnqueueToWriteLatency(net.spy.memcached.ops.Operation op, long operationAttachedNs) {
+        if (op == null || operationAttachedNs <= 0L) return;
+        try {
+            final long wc = op.getWriteCompleteTimestamp();
+            if (wc <= 0L || wc < operationAttachedNs) return;
+            loopEnqueueToWriteLatency.record(wc - operationAttachedNs, TimeUnit.NANOSECONDS);
+        } catch (Throwable t) {
+            if (log.isDebugEnabled()) log.debug("recordLoopEnqueueToWriteLatency failed", t);
```
akhaku: we don't really need the `isDebugEnabled` check here right since the expression in log.debug is cheap

akhaku (approved):
technically not sent on the wire but passed to the non-blocking socket, right?

Generally LGTM

evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheClient.java
jasonk000: Yes. Updated the comments to make that more clear, and dropped that catch there since the onyl thing we have is a `record()` call. -> https://github.com/Netflix/EVCache/pull/193/commits/a6d34e2295357b0ed65ae4dd3a55e18e30bd5e51

jasonk000:
I think this needs a little work. Even though per-op latency is OK, when we multiply it by 100x for a 100-element bulk call the latency can stack up to 10's of microseconds. Not a lot, but, if we can reorder the PR i think we can do a bit better.

(TODO note for myself - [this code](https://github.com/Netflix/EVCache/pull/193/changes#diff-fa4afd745b485d8637737baa57e8933888569680210032fc4dd5d23b69d6ae8fR386) should, instead of writing directly to metrics, accumulate metrics in a [TimerBatchUpdater](https://github.com/Netflix/spectator/blob/main/spectator-api/src/main/java/com/netflix/spectator/api/TimerBatchUpdater.java)). Likely this batching can be applied to other evcache batch query metrics too.

PR #196, opened 2026-06-04
EVCacheValue: opt-in compact binary serialization with backwards-compatible reads

joegoogle123:
## Summary
Hashed-key values are wrapped in an `EVCacheValue` envelope `(key, value, flags, ttl, createTime)` that is currently serialized with Java `ObjectOutputStream`, adding ~50–80 bytes of structural overhead per item. This adds a compact, length-prefixed binary format for the envelope while remaining fully backwards-compatible on reads.

This PR is **two commits intended to be squashed at merge**:
1. `EVCacheValue: opt-in compact binary serde with backwards-compatible read` — the wire format, transcoder dispatch, and the FP-gated rollout switch.
2. `Route the EVCacheValue binary-serialization toggle through EVCacheTranscoderProperties` — extracts the FP resolution into a typed bundle with a per-app → global → static-default chain.

## What changed
- **New `EVCacheValueSerde`** class (`com.netflix.evcache.pool`) — public-final-non-instantiable codec, owns the wire format and all error handling:
  - `static byte[] serialize(EVCacheValue)` — length-prefixed binary layout: `[magic 0x0C][reserved 0x00][int keyLen][key UTF-8][int valLen][value][int flags][long ttl][long createTime]`.
  - `static EVCacheValue deserialize(byte[])` — bounds-checks length prefixes before allocating; on any corruption / unexpected exception **warn-logs the failing field and a truncated hex dump (via Apache Commons `Hex.encodeHexString`, capped at 1024 bytes) and returns `null`**. Matches `BaseSerializingTranscoder`'s resilience contract (corruption → cache miss → caller refills from source of truth) so a single corrupt entry never crashes a get / getBulk / async pipeline.
  - `static boolean isBinaryFormat(byte[])` — exposed for the dispatcher.
- **`EVCacheTranscoder`** becomes a thin dispatcher (no try/catch):
  - `serialize`: gates on `properties.isBinarySerializationEnabled() && o instanceof EVCacheValue` → `EVCacheValueSerde.serialize`; else `super.serialize` (Java).
  - `deserialize`: dispatches on `EVCacheValueSerde.isBinaryFormat` → `EVCacheValueSerde.deserialize`; else `super.deserialize`.
- **`EVCacheValue`** stays a pure POJO (codec moved out; constructor unchanged from pre-PR).
- **`EVCacheTranscoderProperties`** (new bundle in `com.netflix.evcache.config`) — typed access to the FPs that govern `EVCacheTranscoder`. Today it holds one entry (`USE_BINARY_SERIALIZATION`); the `Key` enum is the extension point for future transcoder properties. Resolution chain per property: per-app override → global default → static default, read once at construction and cached. A `getProperty(Key, Class, default)` escape hatch supports future dynamic reads without forcing every caller off the cached primitive.
- **`EVCacheImpl`** constructs an `EVCacheTranscoderProperties` for the envelope transcoder and injects it; max size is still read inline; compression stays pinned at `Integer.MAX_VALUE` so the leading magic byte is never gzip-rewritten.
- Reads auto-detect format by the leading byte (`0xAC 0xED` = legacy Java, `0x0C` = binary), so a new client decodes existing cache entries unchanged.

## Format-flag decision (reuse `SERIALIZED` + magic byte, not a fresh flag)
The binary envelope keeps the existing `SERIALIZED` flag and is disambiguated from Java by the leading byte, rather than allocating a new `CachedData` flag. Rationale:
- `SERIALIZED` semantically still means "serialized object → `deserialize()`"; the codec choice (Java vs binary) lives *inside* `deserialize()`. No flag constant is reassigned or repurposed, and `decode()` branch order is untouched.
- Consumers that route on `SERIALIZED` (e.g. the admin inspector, cache-warmer) keep working without a new flag constant to propagate.
- An old reader that hits binary bytes under `SERIALIZED` throws `StreamCorruptedException` (fails loud) rather than silently decoding garbage — which a fresh low-byte flag would cause (`decodeString`) on old readers.
- Invariant (documented in `EVCacheValueSerde` Javadoc): `SERIALIZED` payloads are self-describing by leading byte; a future third format must use a distinct non-colliding magic + the reserved version byte.

## Reserved version byte
Byte index 1 of the binary payload is reserved (always `0x00` today). Reader read-and-ignores; not validated. Reason: forward-compat without an emergency reader rollout. If today's readers rejected any non-zero version, introducing a v2 in the future would require shipping reader support fleet-wide *before* any writer could emit a v2 byte, and a single misconfigured writer would crash all readers. By accepting any value today, future readers can branch on this byte to introduce breaking format changes backwards-compatibly.

## Feature Property (rollout gate)
Resolved through `EVCacheTranscoderProperties` with a three-level fallback:
1. **Per-app override:** `<appName>.binary.serialization.enabled`
2. **Global default:** `default.evcache.binary.serialization.enabled`
3. **Static default:** `false`

- **Read once at client construction** and cached as a primitive on the (immutable) transcoder ⇒ **deploy/restart required to take effect; this is NOT a live runtime toggle.** Flip the property, then redeploy the consuming app.
- Default-off means production keeps writing Java; reads auto-detect both formats. Roll out **reader-first**: ship this change everywhere — including the admin inspector and cache-warmer — before enabling the FP for any writer.

## Compatibility
- A client with this change decodes existing Java-serialized values unchanged (dual-format read).
- With the FP off (default), wire output is byte-identical to today.
- Corrupt binary payloads degrade to cache miss (null), matching the existing Java path. A single corrupt entry never crashes the caller.

## Testing
`EVCacheValueSerdeTest` — 17 cases via the public `EVCacheTranscoder.encode`/`decode` API:
- Binary round-trip across edge cases (empty / unicode key / large 2 MB value / negative flags / zero ttl / negative & MAX createTime / MIN flags)
- Transcoder wire shape (binary flag on): SERIALIZED flag set, leading byte `0x0C`, reserved byte `0x00`
- Default transcoder (flag OFF) writes Java (`0xAC 0xED`) but reads both formats
- Backwards-compat: legacy `ObjectOutputStream`-serialized envelope still decodes
- Non-`EVCacheValue` passthrough (ArrayList stays on the Java path even with binary flag on)
- Size win: `testBinaryIsSmallerThanJava` asserts the binary envelope is strictly smaller than the Java one for a representative item
- Malformed binary: truncated, bogus oversize keyLen, negative keyLen all decode to null (logged with field + hex dump)
- Forward-compat: `testDecodeBinaryWithOptionalAdditiveFieldIsForwardCompat` — appends an unknown trailing field to a typical payload and asserts the reader still returns the typical envelope (cross-consistency check: if a real new field is added everywhere, the test continues to pass; if someone adds a non-additive field, this test fails).

`EVCacheTranscoderPropertiesTest` — 5 cases for the resolution chain: per-app wins, global fallback when per-app unset, static default when both unset, per-app beats global, null appName routes to global.

`./gradlew :evcache-core:test` green. Also smoke-tested end-to-end against a locally-published snapshot in a sample Spring Boot Netflix app — round-trip put/get/delete on hashed keys passed.

Chunked-payload integration is not covered by an automated test in this PR — chunking lives in `EVCacheClient.createChunks`/`assembleChunks`, which are content-opaque (byte copy + CRC + manifest) and require a live client to exercise. The binary format introduces no new chunking risk by construction: `assembleChunks` reassembles bytes byte-for-byte and CRC-checks them against the manifest before handing the result to the transcoder.

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
@@ -6,6 +6,7 @@ import com.netflix.archaius.api.Property;
 import com.netflix.archaius.api.PropertyRepository;
 import com.netflix.evcache.EVCacheInMemoryCache.DataNotFoundException;
 import com.netflix.evcache.EVCacheLatch.Policy;
+import com.netflix.evcache.config.EVCacheTranscoderProperties;
 import com.netflix.evcache.dto.KeyMapDto;
 import com.netflix.evcache.event.EVCacheEvent;
 import com.netflix.evcache.event.EVCacheEventListener;
@@ -76,6 +77,7 @@ import rx.Single;
 public class EVCacheImpl implements EVCache, EVCacheImplMBean {
 
     private static final Logger log = LoggerFactory.getLogger(EVCacheImpl.class);
+    private static final int COMPRESSION_THRESHOLD_BYTES = Integer.MAX_VALUE;
 
     private final Clock clock;
     private final String _appName;
@@ -164,9 +166,9 @@ public class EVCacheImpl implements EVCache, EVCacheImplMBean {
         this.maxHashLength = propertyRepository.get(appName + ".max.hash.length", Integer.class).orElse(-1);
         this.encoderBase = propertyRepository.get(appName + ".hash.encoder", String.class).orElse("base64");
         this.autoHashKeys = propertyRepository.get(_appName + ".auto.hash.keys", Boolean.class).orElseGet("evcache.auto.hash.keys").orElse(false);
-        this.evcacheValueTranscoder = new EVCacheTranscoder();
-        evcacheValueTranscoder.setCompressionThreshold(Integer.MAX_VALUE);
-
+        final EVCacheTranscoderProperties evCacheTranscoderProperties = new EVCacheTranscoderProperties(_appName, propertyRepository);
+        this.evcacheValueTranscoder = new EVCacheTranscoder(evCacheTranscoderProperties);
+        evcacheValueTranscoder.setCompressionThreshold(COMPRESSION_THRESHOLD_BYTES);
         // default max key length is 200, instead of using what is defined in MemcachedClientIF.MAX_KEY_LENGTH (250). This is to accommodate
         // auto key prepend with appname for duet feature.
         this.maxKeyLength = propertyRepository.get(_appName + ".max.key.length", Integer.class).orElseGet("evcache.max.key.length").orElse(200);
--- a/evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheValueSerde.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheValueSerde.java
@@ -1,0 +1,196 @@
+package com.netflix.evcache.pool;
+
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+import org.apache.commons.codec.binary.Hex;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Length-prefixed binary wire format for the {@link EVCacheValue} envelope. EVCache wraps a
+ * value in an {@code EVCacheValue} when the canonical key has to be hashed (see
+ * {@code EVCacheImpl.getEVCacheKey}) so the pre-hash key is preserved for collision detection.
+ *
+ * <pre>
+ * [byte 0: magic 0x0C][byte 1: reserved/version 0x00]
+ * [int keyLen][key UTF-8 bytes]
+ * [int valLen][value bytes]
+ * [int flags][long ttl][long createTime]
+ * [... optional extension fields appended by newer writers ...]
+ * </pre>
+ *
+ * <ul>
+ *   <li><b>Magic {@code 0x0C}</b> disambiguates from Java {@code ObjectOutputStream} (starts
+ *       {@code 0xAC 0xED}); callers route via {@link #isBinaryFormat(byte[])}.</li>
+ *   <li><b>Reserved/version byte</b> is currently {@code 0x00}, read-and-ignored. Bump only
+ *       for breaking changes (see Upgrades).</li>
+ *   <li><b>End of envelope</b> is implicit at {@code bytes.length}. There is no declared body
+ *       length on the wire; bytes past the last known field are treated as extension data for
+ *       additive forward-compat (see Upgrades).</li>
+ *   <li><b>Byte order:</b> big-endian / network, set explicitly on both sides.</li>
+ *   <li><b>Error contract:</b> any corrupt/truncated input returns {@code null} after a WARN
+ *       log identifying the failing field and a (truncated) hex dump of the bytes. Matches
+ *       {@code BaseSerializingTranscoder}'s resilience contract — caller sees a cache miss.</li>
+ * </ul>
+ *
+ * <h2>Upgrades</h2>
+ *
+ * <p><b>Additive optional (non-breaking).</b> Append a new field at the end of the envelope,
+ * after {@code createTime}. Older readers stop after the known fields and never look at the
+ * extension bytes. Newer readers MUST gate each added field with {@code buffer.hasRemaining()}
+ * and supply a default when absent — they will encounter items written by old writers
+ * (in cache until TTL expires) that don't contain the field. Only works when a graceful
+ * default exists. A new <i>required</i> field has no acceptable default and is therefore
+ * Breaking, not additive.
+ *
+ * <p><b>Breaking</b> (field reorder, type widen, semantic change, new required field):
+ * rollout MUST be <i>reader-before-writer</i> — items written by an early writer would be
+ * silently misparsed by lagging readers and survive until TTL.
+ * <ol>
+ *   <li>Ship a version-aware reader that branches on byte 1: {@code 0x00} stays on the current
+ *       decoder, the new value routes to the new decoder, unknown values
+ *       {@link #logCorruption(byte[], String)} and return {@code null}. Deploy to 100% of every
+ *       consumer that calls {@link #deserialize} (clients, admin tools, cache warmers,
+ *       replicators).</li>
+ *   <li>Wait for the longer of (full reader rollout) and (max item TTL).</li>
+ *   <li>Then ship the new writer gated by a per-app FastProperty so canary is possible.</li>
+ *   <li>Never reuse a version byte value for a different layout.</li>
+ *   <li>Keep the old decoder path indefinitely — items live until their TTL expires.</li>
+ * </ol>
+ */
+public final class EVCacheValueSerde {
+
+    private static final Logger log = LoggerFactory.getLogger(EVCacheValueSerde.class);
+
+    static final byte BINARY_SERDE_MAGIC_CONSTANT_BYTE = 0x0C; // 12
+    private static final byte RESERVED_VERSION_BYTE = 0x00;
+
+    private static final int CORRUPT_PAYLOAD_LOG_LIMIT = 1024;
+
+    private EVCacheValueSerde() {
+        // Utility class; not instantiable.
+    }
+
+    /** True iff {@code bytes} starts with the binary envelope magic byte. */
+    public static boolean isBinaryFormat(byte[] bytes) {
+        return bytes != null && bytes.length > 0 && bytes[0] == BINARY_SERDE_MAGIC_CONSTANT_BYTE;
+    }
+
+    /**
+     * Encode an {@link EVCacheValue} into its compact binary envelope. Key and value must be
+     * non-null — the {@link com.netflix.evcache.EVCacheTranscoder} / {@code CachedData} pipeline
+     * above already rejects nulls.
+     */
+    public static byte[] serialize(EVCacheValue v) {
+        final byte[] keyBytes = v.getKey().getBytes(StandardCharsets.UTF_8);
+        final byte[] valueBytes = v.getValue();
+
+        final int bufferSize =
+                Byte.BYTES + Byte.BYTES                // magic + reserved/version
+              + Integer.BYTES + keyBytes.length        // keyLen + key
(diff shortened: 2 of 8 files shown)


evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheValueSerde.java
```diff
@@ -0,0 +1,189 @@
...
+                return null;
+            }
+            field = "key";
+            final byte[] keyBytes = new byte[keyLength];
+            buffer.get(keyBytes);
+            final String key = new String(keyBytes, StandardCharsets.UTF_8);
+
+            field = "valueLength";
+            final int valueLength = buffer.getInt();
+            if (valueLength < 0 || valueLength > buffer.remaining()) {
```
bihaoxwork, line 147: How do we handle the case the keyLength, valueLength are corrupted less? It seems like we will finish with a incorrect data.
Should we check `if (buffer.hasRemaining()) return null;` after read the buffer?

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
```diff
@@ -77,6 +77,12 @@ public class EVCacheImpl implements EVCache, EVCacheImplMBean {
 
     private static final Logger log = LoggerFactory.getLogger(EVCacheImpl.class);
 
+    // The envelope transcoder used for hashed-key EVCacheValue wrapping must NOT compress its output:
+    // reads detect the format by the leading byte (0xAC = legacy Java, 0x0C = compact binary), which
+    // gzip would mask (compressed payloads start with 0x1F 0x8B). Disable by setting the threshold
+    // higher than any plausible value size.
```
bihaoxwork: Could you explain this a bit? My understanding is that we will first decompress, then decide whether we will de-serialization?

joegoogle123: Hmm I think this comment can be removed, if you look closely at the diff, what you will notice is previously we were __disabling__  compression. This appears to be a post-hoc hallucination from claude.

evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheValueSerde.java
bihaoxwork: **`valueLength` corrupted *smaller* → corrupt data returned as a hit, not `null`.**

The check `valueLength > buffer.remaining()` still passes, so we read `flags`/`ttl`/`createTime` out of value bytes and return a **non-null** `EVCacheValue` with no exception. `key` is read before `valueLength`, so it stays intact and the collision check passes — the caller gets corrupt data as a cache *hit* instead of the documented `null`.

Example — `EVCacheValue(key="ab", value="WXYZ", flags=1, ttl=2, createTime=3)`:

```
serialized (36B):
0C 00 | 00 00 00 02 61 62 | 00 00 00 04 57 58 59 5A | 00 00 00 01 | <ttl=2, 8B> | <createTime=3, 8B>
       | keyLen=2  "ab"   | valLen=4    "WXYZ"      |  flags=1     |

flip one byte, valLen 4 -> 0:
0C 00 | 00 00 00 02 61 62 | 00 00 00 00 57 58 59 5A | 00 00 00 01 | ...
                            ^^^^^^^^^^^ valLen=0
```

`deserialize()` on the corrupted bytes today returns (no exception):

```
EVCacheValue{key="ab", value=[], flags=0x5758595A, ttl=4294967296, createTime=8589934592}
```

`key` is still `"ab"` so the collision check passes; `value` is empty, and the real value bytes `WXYZ` (`0x57 0x58 0x59 0x5A`) got reinterpreted as `flags`. Returned as a hit.

Rejecting leftover bytes after the last field fixes it:

```java
final long createTime = buffer.getLong();
if (buffer.hasRemaining()) return null;   // a corrupted length left bytes unconsumed
return new EVCacheValue(key, valueBytes, flags, ttl, createTime);
```

On the example this trips `hasRemaining()` (4 leftover bytes) → `null` (cache miss). A *larger* `valueLength` is already safe (it over-reads → `BufferUnderflowException` → null); only the *smaller* case slips through.

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
bihaoxwork: You're right, compression was already disabled pre-PR. We can just remove the comments

evcache-core/src/main/java/com/netflix/evcache/pool/EVCacheValueSerde.java
joegoogle123: If we want to support adding new fields we will can't do a `buffer.hasRemaining()` check.  Instead I added `totalLength` field into wire format. Which will let us check that we read all the bytes that are expected to be read.  This will be able to guard against corruption if the `keyLength` or `valueLength` is wrong.

joegoogle123: After trying out the `bodyLength` change I realized it can't support __both__ adding new optional fields and guarding against bit corruption in `keyLength` or `bodyLength`. Given Java object serialization, what we were doing before, suffers from the same issue, I think it's ok to ship this change __without__  addressing the concern around bit corruption.

bihaoxwork (approved):

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
```diff
@@ -164,8 +166,12 @@ public class EVCacheImpl implements EVCache, EVCacheImplMBean {
...
         this.encoderBase = propertyRepository.get(appName + ".hash.encoder", String.class).orElse("base64");
         this.autoHashKeys = propertyRepository.get(_appName + ".auto.hash.keys", Boolean.class).orElseGet("evcache.auto.hash.keys").orElse(false);
-        this.evcacheValueTranscoder = new EVCacheTranscoder();
-        evcacheValueTranscoder.setCompressionThreshold(Integer.MAX_VALUE);
+        // Whether the EVCacheValue envelope (hashed keys) is written using the compact binary format
+        // instead of native Java serialization.
+        final boolean useBinarySerialization = propertyRepository.get(_appName + ".envelope.binary.serialization.enabled", Boolean.class)
+                .orElseGet("evcache.envelope.binary.serialization.enabled").orElse(false).get();
+        final int maxValueSize = propertyRepository.get("default.evcache.max.data.size", Integer.class).orElse(20 * 1024 * 1024).get();
+        this.evcacheValueTranscoder = new EVCacheTranscoder(maxValueSize, ENVELOPE_COMPRESSION_DISABLED, useBinarySerialization);
```
shy-1234: is it intentional to change the compression threshold from Integer.MAX_VALUE to default.evcache.max.data.size?

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
```diff
@@ -164,8 +166,12 @@ public class EVCacheImpl implements EVCache, EVCacheImplMBean {
         this.maxHashLength = propertyRepository.get(appName + ".max.hash.length", Integer.class).orElse(-1);
         this.encoderBase = propertyRepository.get(appName + ".hash.encoder", String.class).orElse("base64");
         this.autoHashKeys = propertyRepository.get(_appName + ".auto.hash.keys", Boolean.class).orElseGet("evcache.auto.hash.keys").orElse(false);
-        this.evcacheValueTranscoder = new EVCacheTranscoder();
-        evcacheValueTranscoder.setCompressionThreshold(Integer.MAX_VALUE);
+        // Whether the EVCacheValue envelope (hashed keys) is written using the compact binary format
+        // instead of native Java serialization.
+        final boolean useBinarySerialization = propertyRepository.get(_appName + ".envelope.binary.serialization.enabled", Boolean.class)
```
shy-1234: nit: should we get this property also in EVCacheTranscoder like the other 2 properties there?

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
shy-1234: sorry let me rephrase. I mean the missing evcacheValueTranscoder.setCompressionThreshold(Integer.MAX_VALUE);

it might be by mistake in the past since I also saw default.evcache.compression.threshold in the EVCacheTranscoder constructor. But just wanted to raise that after this change, there can be a behavior difference to the existing clusters. (changing from INT_MAX to the default 120)

shy-1234 (approved):
maybe just give a second look to the compression threshold thing. Overall LGTM!

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
joegoogle123: `setCompressionThreshold` is called in the constructor. I think this should preserve the existing behavior
```
this.evcacheValueTranscoder = new EVCacheTranscoder(maxValueSize, ENVELOPE_COMPRESSION_DISABLED, useBinarySerialization);
```
Can you elaborate on how we could erroneously go from `Integer.MAX_VALUE`?

shy-1234: Here is the [code](https://github.com/Netflix/EVCache/blob/ae387b27de81540d3e8f79860e64fd8537e534d1/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java#L168)

shy-1234: I might have mis-read the code... One last nit: the naming ENVELOPE_COMPRESSION_DISABLED seems a bit misleading...? (sounds like a boolean)

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
joegoogle123: Created new java object `EVCacheTranscoderProperties` to hold static and dynamic properties for the transcoder. Think it will play nicely with the compression [PR](https://github.com/Netflix/EVCache/pull/197).

evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
joegoogle123: updated to `COMPRESSION_THRESHOLD_BYTES`.

PR #197, opened 2026-06-10
Add Zstd support for EVCache

janewang1680:
**Summary**
Add pluggable compression algorithm support (gzip, zstd) via a new CompressionAlgorithm enum on EVCacheSerializingTranscoder, with magic-byte auto-detection on decode so existing gzip-compressed cache entries
continue to decode correctly. Algorithm and zstd level are configurable via two new fast properties:
default.evcache.compression.algorithm (default GZIP)
default.evcache.compression.zstd.level (default 3)
Fix a pre-existing bug in the compression ratio metric where compressed.length / b.length always evaluated to 1.0 because b was reassigned to compressed before the ratio calculation. The metric now uses the
captured originalLength, producing accurate ratios.
**Test plan**
 ./gradlew :evcache-core:test --tests "com.netflix.evcache.EVCacheSerializingTranscoderTest" — 17 tests covering enum values, default constructor, gzip/zstd round-trip, cross-decode (gzip-written read by zstd
transcoder and vice versa), and FP wiring.
 Verify backward compatibility: data written with gzip in production decodes correctly after deploy.
 Verify compression ratio metric (COMPRESSION_RATIO timer) reports realistic ratios (not always ~100%) in a canary.
Verify it works in our internal services

### DIFF
--- a/evcache-core/build.gradle
+++ b/evcache-core/build.gradle
@@ -43,7 +43,7 @@ dependencies {
         api group:"joda-time",                    name:"joda-time",                        version:"latest.release"
         api group:"javax.annotation",             name:"javax.annotation-api",             version:"latest.release"
         api group:"com.github.fzakaria",          name:"ascii85",                          version:"latest.release"
-        api group:"com.github.luben",             name:"zstd-jni",                         version:"latest.release"
+        api group:"com.github.luben",             name:"zstd-jni",                         version:"1.5.7-11"
 
         testImplementation group:"org.testng",               name:"testng",                           version:"7.5"
         testImplementation group:"com.beust",                name:"jcommander",                       version:"1.72"
--- a/evcache-core/build.gradle
+++ b/evcache-core/build.gradle
@@ -43,6 +43,7 @@ dependencies {
         api group:"joda-time",                    name:"joda-time",                        version:"latest.release"
         api group:"javax.annotation",             name:"javax.annotation-api",             version:"latest.release"
         api group:"com.github.fzakaria",          name:"ascii85",                          version:"latest.release"
+        api group:"com.github.luben",             name:"zstd-jni",                         version:"latest.release"
 
         testImplementation group:"org.testng",               name:"testng",                           version:"7.5"
         testImplementation group:"com.beust",                name:"jcommander",                       version:"1.72"
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
@@ -29,8 +29,6 @@ import com.netflix.evcache.metrics.EVCacheMetricsFactory;
 import com.netflix.spectator.api.BasicTag;
 import com.netflix.spectator.api.Tag;
 import net.spy.memcached.CachedData;
-import net.spy.memcached.compat.log.LoggerFactory;
-import net.spy.memcached.compat.log.Logger;
 import net.spy.memcached.transcoders.BaseSerializingTranscoder;
 import net.spy.memcached.transcoders.Transcoder;
 import net.spy.memcached.transcoders.TranscoderUtils;
@@ -45,6 +43,8 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Transcoder that serializes and compresses objects.
@@ -233,10 +233,10 @@ public class EVCacheSerializingTranscoder extends BaseSerializingTranscoder impl
         switch (compressionAlgorithm) {
             case ZSTD:
                 int zstdLevel = zstdLevelProperty.orElse(DEFAULT_ZSTD_COMPRESSION_LEVEL).get();
-                logger.info("algorithm: " + compressionAlgorithm + ", level: " + zstdLevel + ", appName: " + appName);
+                logger.debug("algorithm: " + compressionAlgorithm + ", level: " + zstdLevel + ", appName: " + appName);
                 return Zstd.compress(in, zstdLevel);
             case GZIP:
-                logger.info("algorithm: " + compressionAlgorithm + ", appName:" + appName);
+                logger.debug("algorithm: " + compressionAlgorithm + ", appName:" + appName);
                 return super.compress(in);
             default:
                 throw new IllegalArgumentException("Unsupported compression algorithm: " + compressionAlgorithm);
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
@@ -24,8 +24,10 @@ package com.netflix.evcache;
 
 import com.github.luben.zstd.Zstd;
 import com.github.luben.zstd.ZstdInputStream;
-import com.netflix.archaius.api.Property;
+import com.netflix.evcache.config.EVCacheTranscoderProperties;
 import com.netflix.evcache.metrics.EVCacheMetricsFactory;
+import com.netflix.evcache.pool.ServerGroup;
+import com.netflix.evcache.util.EVCacheConfig;
 import com.netflix.spectator.api.BasicTag;
 import com.netflix.spectator.api.DistributionSummary;
 import com.netflix.spectator.api.Tag;
@@ -82,6 +84,8 @@ public class EVCacheSerializingTranscoder extends BaseSerializingTranscoder impl
     protected final String appName;
     private final EnumMap<CompressionAlgorithm, DistributionSummary> compressionRatioSummaries;
 
+    protected final EVCacheTranscoderProperties properties;
+
     /**
      * Get a serializing transcoder with the default max data size.
      */
@@ -90,18 +94,23 @@ public class EVCacheSerializingTranscoder extends BaseSerializingTranscoder impl
     }
 
     /**
-     * Get a serializing transcoder that specifies the max data size.
+     * Get a serializing transcoder that specifies the max data size. Builds a default
+     * {@link EVCacheTranscoderProperties} bundle from
+     * {@link EVCacheConfig#getInstance()} — subclasses/callers that want per-app
+     * resolution should use {@link #EVCacheSerializingTranscoder(EVCacheTranscoderProperties, int)}.
      */
     public EVCacheSerializingTranscoder(int max) {
-        this(null, max);
+        this(new EVCacheTranscoderProperties(null, EVCacheConfig.getInstance().getPropertyRepository()), max);
     }
 
     /**
-     * Get a serializing transcoder that specifies the owning app name and the max data size.
+     * Get a serializing transcoder with the supplied transcoder-property bundle. The bundle is
+     * exposed to subclasses via {@link #properties} so downstream transcoders can consult the
+     * same three-level (per-app → global → static default) resolution chain.
      */
-    public EVCacheSerializingTranscoder(String appName, int max) {
+    public EVCacheSerializingTranscoder(EVCacheTranscoderProperties properties, int max) {
         super(max);
-        this.appName = appName;
+        this.properties = properties;
         this.compressionRatioSummaries = buildCompressionRatioSummaries(appName);
     }
 
@@ -116,14 +125,6 @@ public class EVCacheSerializingTranscoder extends BaseSerializingTranscoder impl
             summaries.put(algo, EVCacheMetricsFactory.getInstance().getDistributionSummary(EVCacheMetricsFactory.COMPRESSION_RATIO, tagList));
         }
         return summaries;
-    }
-
-    public void setCompressionAlgorithmProperty(Property<String> algorithmProperty) {
-        this.compressionAlgorithmProperty = algorithmProperty;
-    }
-
-    public void setCompressionLevelProperty(Property<Integer> levelProperty) {
-        this.zstdLevelProperty = levelProperty;
     }
 
     @Override
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
@@ -233,10 +233,10 @@ public class EVCacheSerializingTranscoder extends BaseSerializingTranscoder impl
         switch (compressionAlgorithm) {
             case ZSTD:
                 int zstdLevel = zstdLevelProperty.orElse(DEFAULT_ZSTD_COMPRESSION_LEVEL).get();
-                logger.error("!!!!!!!!!! algoritthm: " + compressionAlgorithm + ", level: " + zstdLevel + ", appName: " + appName);
+                logger.info("algorithm: " + compressionAlgorithm + ", level: " + zstdLevel + ", appName: " + appName);
                 return Zstd.compress(in, zstdLevel);
             case GZIP:
-                logger.error("!!!!!!!!!! algoritthm: " + compressionAlgorithm + ", appName:" + appName);
+                logger.info("algorithm: " + compressionAlgorithm + ", appName:" + appName);
                 return super.compress(in);
             default:
                 throw new IllegalArgumentException("Unsupported compression algorithm: " + compressionAlgorithm);
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
@@ -241,11 +241,11 @@ public class EVCacheSerializingTranscoder extends BaseSerializingTranscoder impl
         switch (compressionAlgorithm) {
             case ZSTD:
                 int zstdLevel = transcoderProperties.getZstdCompressionLevelProperty().get();
-                getLogger().debug("algorithm: {}, level: {}, appName: {}", compressionAlgorithm, zstdLevel, appName);
+                getLogger().debug("algorithm: %s, level: %d, appName: %s", compressionAlgorithm, zstdLevel, appName);
                 compressed = Zstd.compress(in, zstdLevel);
                 break;
             case GZIP:
-                getLogger().debug("algorithm: {}, appName: {}", compressionAlgorithm, appName);
+                getLogger().debug("algorithm: %s, appName: %s", compressionAlgorithm, appName);
                 compressed = super.compress(in);
                 break;
             default:
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
@@ -37,6 +37,8 @@ import net.spy.memcached.CachedData;
 import net.spy.memcached.transcoders.BaseSerializingTranscoder;
 import net.spy.memcached.transcoders.Transcoder;
 import net.spy.memcached.transcoders.TranscoderUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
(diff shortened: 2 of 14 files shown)


evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
```diff
@@ -63,10 +68,16 @@ public class EVCacheSerializingTranscoder extends BaseSerializingTranscoder impl
...
+    public enum CompressionAlgorithm { GZIP, ZSTD }
+
+    public static final int DEFAULT_ZSTD_COMPRESSION_LEVEL = 3;
+
+    private static final int ZSTD_MAGIC = 0xFD2FB528;
 
     private final TranscoderUtils tu = new TranscoderUtils(true);
-    private Timer timer;
+    private Property<String> compressionAlgorithmProperty;
+    private Property<Integer> zstdLevelProperty;
```
srrangarajan: These are never initialized in the base constructors, so any caller who does new EVCacheSerializingTranscoder(max) directly and then encodes something large enough to compress will NPE here. 

Before this PR that code path just fell through to BaseSerializingTranscoder.compress() and worked fine — this is a regression. Can we initialize these to constant defaults (GZIP /                      DEFAULT_ZSTD_COMPRESSION_LEVEL) in the field declaration so the class is always safe to use standalone, even without calling the setters?

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
```diff
@@ -179,31 +206,103 @@ public CachedData encode(Object o) {
...
+        if (data == null || data.length < 4) return false;
+        int magic = ByteBuffer.wrap(data, 0, 4).order(ByteOrder.LITTLE_ENDIAN).getInt();
+        return magic == ZSTD_MAGIC;
+    }
 
-        timer.record(ratio_percentage, TimeUnit.MILLISECONDS);
+    private byte[] decompressZstd(byte[] in) {
+        long originalSize = Zstd.getFrameContentSize(in);
+        if (originalSize > Integer.MAX_VALUE) {
+            getLogger().warn("Zstd decompressed size exceeds int range: " + originalSize);
```
srrangarajan: return null, which propagates back through decompress() into decode(). For plain string values, decode() ends up calling decodeString(null) which NPEs. For serialized objects it silently returns null to the caller with no exception — hard to debug. Would prefer throwing a RuntimeException here so the failure is visible rather than swallowed

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
```diff
@@ -179,31 +206,103 @@ public CachedData encode(Object o) {
...
                 flags |= COMPRESSED;
             } else {
                 getLogger().debug("Compression increased the size of %s from %d to %d",
-                        o.getClass().getName(), b.length, compressed.length);
+                        o.getClass().getName(), originalLength, compressed.length);
             }
 
-            long compression_ratio = Math.round((double) compressed.length / b.length * 100);
-            updateTimerWithCompressionRatio(compression_ratio);
+            long ratioPerCent = Math.round((double) compressed.length / originalLength * 100);
```
srrangarajan: thought! recordCompressionRatio calls resolveCompressionAlgorithm() internally, but compress() already called it too. That's two Archaius property lookups per encode for every  compressed value. Minor but this is a hot path. Could we resolve once and pass the algorithm down?

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
```diff
@@ -22,31 +22,36 @@
...
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Transcoder that serializes and compresses objects.
  */
 public class EVCacheSerializingTranscoder extends BaseSerializingTranscoder implements
         Transcoder<Object> {
+    Logger logger = LoggerFactory.getLogger(EVCacheSerializingTranscoder.class);
```
srrangarajan: private static final — without static you get a separate Logger object per transcoder instance

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
```diff
@@ -179,31 +206,103 @@ public CachedData encode(Object o) {
...
+    @Override
+    protected byte[] compress(byte[] in) {
+        if (in == null) throw new NullPointerException("Can't compress null");
+
+        CompressionAlgorithm compressionAlgorithm = resolveCompressionAlgorithm();
+
+        switch (compressionAlgorithm) {
+            case ZSTD:
+                int zstdLevel = zstdLevelProperty.orElse(DEFAULT_ZSTD_COMPRESSION_LEVEL).get();
+                logger.debug("algorithm: " + compressionAlgorithm + ", level: " + zstdLevel + ", appName: " + appName);
```
srrangarajan: minor: use {} instead of +

evcache-core/build.gradle
```diff
@@ -43,6 +43,7 @@ dependencies {
         api group:"joda-time",                    name:"joda-time",                        version:"latest.release"
         api group:"javax.annotation",             name:"javax.annotation-api",             version:"latest.release"
         api group:"com.github.fzakaria",          name:"ascii85",                          version:"latest.release"
+        api group:"com.github.luben",             name:"zstd-jni",                         version:"latest.release"
```
srrangarajan: thought:  latest.release means any nebula upgrade run can silently pull in a new native binary. Worth pinning to a  specific version and upgrading intentionally?

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
janewang1680: updated

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
janewang1680: updated.

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
janewang1680: updated

evcache-core/build.gradle
janewang1680: updated to use  1.5.7.7

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
janewang1680: This metrics was wrong before, always 100%.

Now, I have second thought. We should report the actual compressed size, even if the compressed size is larger. This can help with the tuning the compression threshold. 

Resolved the twice read of FP.

srrangarajan:
@shy-1234 can you also review the PR please

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
```diff
@@ -179,31 +206,106 @@ public CachedData encode(Object o) {
...
+        byte[] buf = new byte[8192];
+        int n;
+        while ((n = in.read(buf)) != -1) {
+            out.write(buf, 0, n);
+        }
+        return out.toByteArray();
     }
 
+    private void recordCompressionRatio(long ratioPerCent, CompressionAlgorithm compressionAlgorithm) {
+        final List<Tag> tagList = new ArrayList<Tag>(2);
```
shy-1234: Should we cache the DistributionSummary instead of object allocation (ArrayList + BasicTags) on every compression? (seems like an unnecessary GC pressure in a hot path)

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
```diff
@@ -179,31 +206,106 @@ public CachedData encode(Object o) {
...
+        }
+        if (originalSize > 0) {
+            // Fast path: frame carries a content-size header (compress() above always does).
+            return Zstd.decompress(in, (int) originalSize);
+        }
+        // Slow path: declared size is 0, unknown (-1), or invalid (-2) — stream-decode and let
+        // ZstdInputStream surface any frame errors.
+        ZstdInputStream zis = null;
+        try {
+             zis = new ZstdInputStream(new ByteArrayInputStream(in));
```
shy-1234, line 319: Based on the above comments, I assume we don't expect to reach this code (slow) path. I suggest we add a metric (or a debug log or even a warn log given that this path seems to indicate some unexpected upstream compress behavior even when it's able to decompress.) here to record any case who reached here.

evcache-core/src/main/java/com/netflix/evcache/connection/BaseAsciiConnectionFactory.java
```diff
@@ -115,7 +115,8 @@ public BlockingQueue<Operation> createWriteOperationQueue() {
     }
 
     public Transcoder<Object> getDefaultTranscoder() {
-        return new EVCacheTranscoder();
+        return new EVCacheTranscoder(appName,
+                client.getPool().getEVCacheClientPoolManager().getEVCacheConfig().getPropertyRepository());
```
shy-1234: nit: why not just appName? seems like there is a constructor that handles this.

evcache-core/src/main/java/com/netflix/evcache/connection/BaseConnectionFactory.java
```diff
@@ -109,7 +109,8 @@ public BlockingQueue<Operation> createWriteOperationQueue() {
     }
 
     public Transcoder<Object> getDefaultTranscoder() {
-        return new EVCacheTranscoder();
+        return new EVCacheTranscoder(appName,
+                client.getPool().getEVCacheClientPoolManager().getEVCacheConfig().getPropertyRepository());
```
shy-1234: same nit: why not just appName? seems like there is a constructor that handles this.

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
janewang1680: cached the taglist. updated the code.

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
janewang1680: added a log.warn()

evcache-core/src/main/java/com/netflix/evcache/connection/BaseAsciiConnectionFactory.java
janewang1680: the constructor without `PropertyRepository config` argumnt uses `EVCacheConfig.getInstance().getPropertyRepository()`, which is a global setting. If an application uses multiple evcaches, they all share the same config, which is NOT optimal, but maybe ok for default.evcache.max.data.size. Ideally, we should have different config for different evache as each may have different requirement. per cache FP should use `client.getPool().getEVCacheClientPoolManager().getEVCacheConfig().getPropertyRepository()`.  

For backward compatibility, we should not change `default.evcache.max.data.size`.

evcache-core/src/main/java/com/netflix/evcache/connection/BaseConnectionFactory.java
janewang1680: the constructor without PropertyRepository config argumnt uses EVCacheConfig.getInstance().getPropertyRepository(), which is a global setting. If an application uses multiple evcaches, they all share the same config, which is NOT optimal, but maybe ok for default.evcache.max.data.size. Ideally, we should have different config for different evache as each may have different requirement. per cache FP should use client.getPool().getEVCacheClientPoolManager().getEVCacheConfig().getPropertyRepository().

For backward compatibility, we should not change default.evcache.max.data.size.

evcache-core/src/main/java/com/netflix/evcache/connection/BaseAsciiConnectionFactory.java
shy-1234: I feel like the statement is not correct, but please correct me if I am wrong.
First of all
```
  EVCacheConfig.propertyRepository is a static field (line 27). There is only ever one value for the entire JVM, regardless of how many EVCacheConfig instances exist.     
  EVCacheConfig is also a singleton with a single INSTANCE.                                                                                                                
                                                                                                                                                                           
  Following the longer path:                                                                                                                                               
  - client.getPool().getEVCacheClientPoolManager() returns the pool manager                                                                                                
  - .getEVCacheConfig() returns this.evcConfig — which is set from line 153: new EVCacheClientPoolManager(..., EVCacheConfig.getInstance())                                
  - .getPropertyRepository() returns the static propertyRepository field                                                                                                   
                                                                                                                                                                           
  Both paths terminate at the exact same static field. They return the same object.
```

Secondly, 
> If an application uses multiple evcaches, they all share the same config, which is NOT optimal

Agree, but IIUC the way the client code achieves per cluster config is by appending evcache cluster name in front of the FP and fallback to the global one (prefix with "evcache") if not provided. (not saying it's good or bad though, but in this case, if your intention is to achieve per cluster config, then i don't think it does the work.)

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
```diff
@@ -179,31 +223,103 @@ public CachedData encode(Object o) {
...
-            final List<Tag> tagList = new ArrayList<Tag>(1);
-            tagList.add(new BasicTag(EVCacheMetricsFactory.COMPRESSION_TYPE, "gzip"));
-            timer = EVCacheMetricsFactory.getInstance().getPercentileTimer(EVCacheMetricsFactory.COMPRESSION_RATIO, tagList, Duration.ofMillis(100));
-        };
+    @Override
+    protected byte[] compress(byte[] in) {
+        if (in == null) throw new NullPointerException("Can't compress null");
+
+        CompressionAlgorithm compressionAlgorithm = compressionAlgorithmProperty == null ? CompressionAlgorithm.GZIP
+                : CompressionAlgorithm.valueOf(compressionAlgorithmProperty.orElse(CompressionAlgorithm.GZIP.name()).get().toUpperCase());
```
joegoogle123: Must the choice of algorithm and the level be dynamic? Could this be something we set at evcache client creation time? This would mean you would need to restart the server to pick up a client change.

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
```diff
@@ -22,31 +22,38 @@
...
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.TimeUnit;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Transcoder that serializes and compresses objects.
  */
 public class EVCacheSerializingTranscoder extends BaseSerializingTranscoder implements
```
joegoogle123, line 58: @janewang1680 
Can we move the compression changes from `EVCacheSerializingTranscoder` down it's subclass `EVCacheTranscoder`.  

Benefits:
1. Isolates most of the compression logic to a single transcoder file.
2. Removes the need for setter methods, default property constants, and null checks `(compressionAlgorithmProperty == null ? ...`, 
3. Keeps `EVCacheSerializingTranscoder` as a generic serialize base class.

evcache-core/src/main/java/com/netflix/evcache/EVCacheTranscoder.java
```diff
@@ -1,32 +1,69 @@
...
+    }
+
+    public EVCacheTranscoder(String appName, PropertyRepository config, int max) {
+        this(appName, config, max, config.get("default.evcache.compression.threshold", Integer.class).orElse(120).get());
+    }
+
+    public EVCacheTranscoder(String appName, PropertyRepository config, int max, int compressionThreshold) {
+        super(appName, max);
+        setCompressionThreshold(compressionThreshold);
+        Property<String> algoProperty = getProperty(config, "evcacheclient.compression.algo", String.class);
```
joegoogle123: Can we name this so the naming convention is uniform?
```
default.evcache.compression.algo
```

evcache-core/src/main/java/com/netflix/evcache/EVCacheTranscoder.java
```diff
@@ -1,32 +1,69 @@
...
+        setCompressionAlgorithmProperty(algoProperty);
+        Property<Integer> zstdLevelProperty = getProperty(config, "evcacheclient.compression.zstd.level", Integer.class);
```
joegoogle123: Same here
```
default.evcache.compression.zstd.level
```

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
janewang1680: I feel it is better to be dynamic. if anything wrong, we can immediately switch back to gzip (this has been very useful when we deployed to our services).

evcache-core/src/main/java/com/netflix/evcache/EVCacheTranscoder.java
janewang1680: sure. updated.

evcache-core/src/main/java/com/netflix/evcache/EVCacheTranscoder.java
janewang1680: updated.

evcache-core/src/main/java/com/netflix/evcache/EVCacheSerializingTranscoder.java
janewang1680: `EVCacheSerializingTranscoder` is directly used in `EVCacheClient`. If we move the compression changes to `EVCacheTranscoder`, `EVCacheTranscoder` will loose the ability to use Zstd for chunk metadata. the metadata and the cache values could be inconsistent regarding compression.

evcache-core/src/main/java/com/netflix/evcache/connection/BaseAsciiConnectionFactory.java
janewang1680: 1. they use different propertyRepository:

client.getPool().getEVCacheClientPoolManager().getEVCacheConfig() resolves to the DI-injected EVCacheConfig:

- EVCacheClientPool.getEVCacheClientPoolManager() returns this.manager (EVCacheClientPool.java:1672-1674), which is the manager passed into its constructor (:134).
- EVCacheClientPoolManager.getEVCacheConfig() returns this.evcConfig (:140-142), and evcConfig is the @Inject constructor parameter (:82-88):

@Inject
public EVCacheClientPoolManager(IConnectionBuilder ..., EVCacheNodeList ..., EVCacheConfig evcConfig) {
    ...
    this.evcConfig = evcConfig;   // the container-provided EVCacheConfig

So in a DI (Guice/Governator) app, evcConfig is the EVCacheConfig that the container built by injecting the application's PropertyRepository. In EVCacheConfig's constructor that hits the repository != null branch (EVCacheConfig.java:48-49):

} else {
    _propertyRepository = repository;   // the app's live, FP-backed repository
}

That app repository is the one wired to the platform's Fast Properties polling, so .get(key).orElse(default).get() re-reads the current value on each call → runtime FP changes are visible.

 The PropertyRepository binding is provided by ArchaiusModule.

2. Sriram has built a RC for me and I have tested in our VHS service, only client.getPool().getEVCacheClientPoolManager().getEVCacheConfig() can pick the change from FP.

joegoogle123 (approved):

srrangarajan:
LGTM

srrangarajan (approved):

shy-1234 (approved):
LGTM. Please also get a +1 from @joegoogle123 and I think we are good to go.

PR #198, opened 2026-06-30
Bundle the transcoder's properties into EVCacheTranscoderProperties

joegoogle123:
## Summary

Route the `EVCacheValue` binary-serialization toggle through a new `EVCacheTranscoderProperties` bundle instead of an inline `propertyRepository.get(...).get()` chain in `EVCacheImpl`. The bundle is a typed extension point — every property it exposes resolves through the same three-level chain, and new transcoder properties land on the `Key` enum without touching call sites.

## What changed

- **New `EVCacheTranscoderProperties`** (`com.netflix.evcache.config`): holds the bundle. `Key` enum carries both the per-app suffix and the global key per property. Constructor `(String appName, PropertyRepository propertyRepository)` snapshots every property at construction. Today's only entry is `USE_BINARY_SERIALIZATION` — additional transcoder properties drop in here.
- **`EVCacheTranscoder`** now has a `(int max, int compressionThreshold, EVCacheTranscoderProperties properties)` ctor that stores the bundle and consults `properties.isBinarySerializationEnabled()` inside `serialize()`. The previous `(int, int, boolean)` ctor is removed; the `(int, int)` ctor builds a bundle with `appName=null` so no-app callers fall through to the global key.
- **`EVCacheImpl`** drops the inline `<app>.envelope.binary.serialization.enabled` read and the `ENVELOPE_COMPRESSION_DISABLED` constant, building the bundle once per app.

## Three-level resolution

Per property, resolved at construction:

1. `<appName>.<appSuffix>` — per-app override
2. `<globalKey>` — fleet-wide global
3. compiled-in static default

For `USE_BINARY_SERIALIZATION`: appSuffix `binary.serialization.enabled`, globalKey `default.evcache.binary.serialization.enabled`, default `false`.

## Property-key rename ⚠️

This PR renames the FastProperty introduced in #196:

| | Before (#196) | After (this PR) |
|---|---|---|
| Per-app | `<app>.envelope.binary.serialization.enabled` | `<app>.binary.serialization.enabled` |
| Global  | `evcache.envelope.binary.serialization.enabled` | `default.evcache.binary.serialization.enabled` |

Rationale: align the global key with the `default.evcache.…` convention used by the other transcoder properties, and drop the `envelope.` prefix that was implicit-already (the transcoder is the envelope codec). Any caller who set the old keys will need to migrate.

## Dynamic-property escape hatch

Future fields that need runtime mutability can skip the cached primitive and call `EVCacheTranscoderProperties#getProperty(Key, Class, defaultValue)` per access — the same three-level chain applies. PR #197's `setCompressionAlgorithmProperty` / `setCompressionLevelProperty` setters on `EVCacheSerializingTranscoder` are orthogonal and untouched here.

## Stacked on

Base is `evcache-value-binary-serde` (PR #196). When #196 merges, this PR's base auto-updates to `master`.

## Test plan

- [x] `./gradlew :evcache-core:test` — green. Five new tests in `EVCacheTranscoderPropertiesTest`:
  - per-app override wins
  - global fallback when per-app unset
  - static default when both unset
  - per-app beats global when both set
  - null appName routes through the global key (for the no-app ctor path)
- [ ] Re-run `joelee-sample-app` smoke test against a republished snapshot. The smoke test's `@TestPropertySource` will need updating to the renamed property keys.

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
@@ -6,6 +6,7 @@ import com.netflix.archaius.api.Property;
 import com.netflix.archaius.api.PropertyRepository;
 import com.netflix.evcache.EVCacheInMemoryCache.DataNotFoundException;
 import com.netflix.evcache.EVCacheLatch.Policy;
+import com.netflix.evcache.config.EVCacheTranscoderProperties;
 import com.netflix.evcache.dto.KeyMapDto;
 import com.netflix.evcache.event.EVCacheEvent;
 import com.netflix.evcache.event.EVCacheEventListener;
@@ -76,8 +77,7 @@ import rx.Single;
 public class EVCacheImpl implements EVCache, EVCacheImplMBean {
 
     private static final Logger log = LoggerFactory.getLogger(EVCacheImpl.class);
-
-    private static final int ENVELOPE_COMPRESSION_DISABLED = Integer.MAX_VALUE;
+    private static final int DEFAULT_COMPRESSION_THRESHOLD = Integer.MAX_VALUE;
 
     private final Clock clock;
     private final String _appName;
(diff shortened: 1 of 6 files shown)


joegoogle123:
Folding the transcoder-properties commit into PR #196; we'll squash at merge.

PR #199, opened 2026-07-02
Route zstd compression config through EVCacheTranscoderProperties bundle (stacked on #197)

joegoogle123:
> **Draft — stacked on top of #197.** Merges in #196 and demonstrates the
> reviewer's suggestion from https://github.com/Netflix/EVCache/pull/197#discussion_r3500593987 —
> compression config can flow through `EVCacheTranscoderProperties` cleanly
> without moving the codec out of `EVCacheSerializingTranscoder`.

## Summary

Combines PR #196 (binary `EVCacheValue` serde + `EVCacheTranscoderProperties`
bundle) with PR #197 (zstd on `EVCacheSerializingTranscoder`) and replaces
PR #197's public setter-injection API with typed `Property<T>` handles owned
by the bundle.

## What changes vs #197

- **Removed** `setCompressionAlgorithmProperty` and `setCompressionLevelProperty`
  from `EVCacheSerializingTranscoder`. Also removed the two mutable
  `Property<T>` fields those setters wrote to.
- **Removed** the redundant `protected final String appName` field on
  `EVCacheSerializingTranscoder` — single source of truth is
  `properties.getAppName()`.
- **Removed** `EVCacheTranscoder(String, PropertyRepository, ...)` constructor
  overloads. Callers pass an `EVCacheTranscoderProperties` instead.
- **Moved** the `CompressionAlgorithm` enum from `EVCacheSerializingTranscoder`
  to `EVCacheTranscoderProperties` so the bundle can expose it as a typed
  return value.
- **Normalized** the compression-config Archaius keys from PR #197's
  `evcacheclient.compression.*` (Java source) to `default.evcache.compression.*`
  (matching PR #197's own PR description and the naming convention already
  used by `default.evcache.max.data.size` / `default.evcache.compression.threshold`).

## The bundle now owns the compression config

New keys and defaults in `EVCacheTranscoderProperties`:

```
COMPRESSION_ALGORITHM   -> per-app: <app>.compression.algorithm
                          global:  default.evcache.compression.algorithm
                          default: CompressionAlgorithm.GZIP

COMPRESSION_ZSTD_LEVEL  -> per-app: <app>.compression.zstd.level
                          global:  default.evcache.compression.zstd.level
                          default: 3
```

The bundle stores two live `Property<T>` handles as final fields, resolved
once at construction:

```java
private final Property<CompressionAlgorithm> compressionAlgorithmProperty;
private final Property<Integer>              zstdCompressionLevelProperty;

// in EVCacheTranscoderProperties ctor:
this.compressionAlgorithmProperty = getProperty(appName, propertyRepository,
        Key.COMPRESSION_ALGORITHM, String.class, DEFAULT_COMPRESSION_ALGORITHM.name())
        .map(s -> CompressionAlgorithm.valueOf(s.toUpperCase()));

this.zstdCompressionLevelProperty = getProperty(appName, propertyRepository,
        Key.COMPRESSION_ZSTD_LEVEL, Integer.class, DEFAULT_COMPRESSION_ZSTD_LEVEL);

public Property<CompressionAlgorithm> getCompressionAlgorithmProperty() {
    return compressionAlgorithmProperty;
}
public Property<Integer> getZstdCompressionLevelProperty() {
    return zstdCompressionLevelProperty;
}
```

The `Property<CompressionAlgorithm>` is built via Archaius's built-in
`Property.map(Function<T, S>)`. `.get()` on the mapped handle re-parses the
underlying String on each call, so **live FP updates propagate without JVM
restart** and callers never see a stale enum. The upper-case wrap keeps the
FP case-insensitive.

## `EVCacheSerializingTranscoder.compress()` becomes trivial

```java
CompressionAlgorithm compressionAlgorithm = properties.getCompressionAlgorithmProperty().get();

switch (compressionAlgorithm) {
    case ZSTD:
        int zstdLevel = properties.getZstdCompressionLevelProperty().get();
        compressed = Zstd.compress(in, zstdLevel);
        break;
    case GZIP:
        compressed = super.compress(in);
        break;
    ...
}
```

No null-checks, no `toUpperCase()` + `valueOf()` boilerplate, no local field
plumbing.

## Testing

`./gradlew :evcache-core:test` — **BUILD SUCCESSFUL**.

- `EVCacheSerializingTranscoderTest` (PR #197's suite): **22 tests pass**,
  rewritten to build transcoders via
  `new EVCacheTranscoderProperties(appName, repo)` instead of the removed
  setters. Tests exercise all four combinations of {global, per-app} × {algo,
  level} keys plus cross-decode via magic-byte detection.
- `EVCacheTranscoderPropertiesTest` (PR #196's suite): 15 tests pass.
- `EVCacheValueSerdeTest` (PR #196's suite): 17 tests pass.

Dynamic-FP behavior end-to-end verified separately on a Spring Boot Netflix
sample app (via `metatron curl -a gate POST /fastproperties/upsert` and observed
propagation into the running JVM's REMOTE Archaius layer within ~30–45s).

## Migration notes for downstream callers

**PR #197 hasn't merged yet** so the blast radius is small — the removed
setters have no external call sites. The renamed Archaius keys
(`default.evcache.compression.*` instead of `evcacheclient.compression.*`)
similarly can't be set in prod yet.

If a downstream caller of `EVCacheTranscoder` was passing
`(appName, PropertyRepository)` (only possible on PR #197's WIP branch),
replace with:

```java
new EVCacheTranscoder(new EVCacheTranscoderProperties(appName, propertyRepository))
```

## Compatibility

Public API on `EVCacheSerializingTranscoder`:
- Removed: `setCompressionAlgorithmProperty(Property<String>)`,
  `setCompressionLevelProperty(Property<Integer>)`.
- Added: `EVCacheSerializingTranscoder(EVCacheTranscoderProperties, int)`.
- Kept: `EVCacheSerializingTranscoder()`, `EVCacheSerializingTranscoder(int max)`.

Public API on `EVCacheTranscoder`:
- Removed: `(String)`, `(String, PropertyRepository)`, `(String, int)`,
  `(String, PropertyRepository, int)`, `(String, int, int)`,
  `(String, PropertyRepository, int, int)` (all PR #197 additions).
- Added: `EVCacheTranscoder(EVCacheTranscoderProperties)`.
- Kept: `()`, `(int max)`, `(int max, int compressionThreshold)`.

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/connection/BaseConnectionFactory.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/connection/BaseConnectionFactory.java
@@ -109,8 +109,8 @@ public class BaseConnectionFactory extends BinaryConnectionFactory {
     }
 
     public Transcoder<Object> getDefaultTranscoder() {
-        return new EVCacheTranscoder(appName,
-                client.getPool().getEVCacheClientPoolManager().getEVCacheConfig().getPropertyRepository());
+        return new EVCacheTranscoder(new com.netflix.evcache.config.EVCacheTranscoderProperties(appName,
+                client.getPool().getEVCacheClientPoolManager().getEVCacheConfig().getPropertyRepository()));
     }
 
     public FailureMode getFailureMode() {
--- a/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/EVCacheImpl.java
@@ -6,6 +6,7 @@ import com.netflix.archaius.api.Property;
 import com.netflix.archaius.api.PropertyRepository;
 import com.netflix.evcache.EVCacheInMemoryCache.DataNotFoundException;
 import com.netflix.evcache.EVCacheLatch.Policy;
+import com.netflix.evcache.config.EVCacheTranscoderProperties;
 import com.netflix.evcache.dto.KeyMapDto;
 import com.netflix.evcache.event.EVCacheEvent;
 import com.netflix.evcache.event.EVCacheEventListener;
@@ -76,6 +77,7 @@ import rx.Single;
 public class EVCacheImpl implements EVCache, EVCacheImplMBean {
 
     private static final Logger log = LoggerFactory.getLogger(EVCacheImpl.class);
+    private static final int COMPRESSION_THRESHOLD_BYTES = Integer.MAX_VALUE;
 
     private final Clock clock;
     private final String _appName;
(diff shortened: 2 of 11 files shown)


evcache-core/src/main/java/com/netflix/evcache/connection/BaseConnectionFactory.java
```diff
@@ -109,8 +109,8 @@ public BlockingQueue<Operation> createWriteOperationQueue() {
     }
 
     public Transcoder<Object> getDefaultTranscoder() {
-        return new EVCacheTranscoder(appName,
-                client.getPool().getEVCacheClientPoolManager().getEVCacheConfig().getPropertyRepository());
+        return new EVCacheTranscoder(new com.netflix.evcache.config.EVCacheTranscoderProperties(appName,
```
janewang1680, line 112: We should one instance of EVCacheTranscoderProperties per appName.

PR #200, opened 2026-07-08
Move binary serialization to EVCacheSerializingTranscoder Refactor

joegoogle123:
## Summary

Follow-up refactor to PR #196. Moves the `EVCacheValue` binary serde dispatch onto the superclass and cleans up the transcoder properties API so callers get typed accessors instead of the generic `getProperty(Key, Class, T)` escape hatch.

Zero behavior change on the wire — the same `serialize`/`deserialize` gating and `EVCacheValueSerde.isBinaryFormat` check just live one class up. This lands ahead of the zstd compression PR (#197) so that PR shows only compression content.

## What changed

- **`EVCacheSerializingTranscoder`** — added `protected` `serialize`/`deserialize` overrides that gate on `transcoderProperties.isBinarySerializationEnabled() && o instanceof EVCacheValue` (moved from `EVCacheTranscoder`). Constructor reordered to `(int max, EVCacheTranscoderProperties)`. New `protected final String appName` field pulled from `properties.getAppName()` for downstream metric tagging. Renamed `properties` → `transcoderProperties`. Removed unused imports (`ServerGroup`, `StringUtils`, `Map`, `ConcurrentHashMap`).
- **`EVCacheTranscoder`** — now a thin wrapper: only `encode()` overrides super (to pass `CachedData` through unchanged). Removed the redundant `asyncDecode`/`decode` overrides that only delegated. Constructors updated to use typed getters and the new super signature.
- **`EVCacheTranscoderProperties`** — snapshotted `maxDataSizeBytes` and `compressionThresholdBytes` as `final int` at construction. Added typed getters `getAppName()`, `getMaxDataSizeBytes()`, `getCompressionThresholdBytes()`. Removed the public `<T> Property<T> getProperty(Key, Class, T)` method — no callers remain now that the transcoder chain uses the typed accessors.
- **`BaseAsciiConnectionFactory` / `BaseConnectionFactory`** — `getDefaultTranscoder()` now passes a per-app `EVCacheTranscoderProperties` bundle so the `<appName>.<key>` prefix reaches the default transcoder path. PR #196 wired this for the value transcoder in `EVCacheImpl` but missed the default transcoder built here.
- **`EVCacheTranscoderPropertiesTest`** — 15 existing resolution-chain cases migrated to the typed getters. Added 2 cases for `getAppName()`.

## Test plan

- [x] `./gradlew :evcache-core:test --tests "com.netflix.evcache.config.EVCacheTranscoderPropertiesTest"` — 17/17 pass
- [x] `./gradlew :evcache-core:test --tests "com.netflix.evcache.pool.EVCacheValueSerdeTest"` — 17/17 pass (dispatch coverage from PR #196 still green since the dispatch just moved up a class)
- [x] `./gradlew :evcache-core:build` — clean, no unused-import warnings

### DIFF
--- a/evcache-core/src/main/java/com/netflix/evcache/config/EVCacheTranscoderProperties.java
+++ b/evcache-core/src/main/java/com/netflix/evcache/config/EVCacheTranscoderProperties.java
@@ -3,8 +3,6 @@ package com.netflix.evcache.config;
 import com.netflix.archaius.api.Property;
 import com.netflix.archaius.api.PropertyRepository;
 
-import static com.netflix.evcache.config.EVCacheTranscoderProperties.Key.BINARY_SERIALIZATION_ENABLED;
-
 /**
  * Properties related to {@link com.netflix.evcache.EVCacheTranscoder}
  * behavior.
@@ -18,8 +16,9 @@ import static com.netflix.evcache.config.EVCacheTranscoderProperties.Key.BINARY_
  * </ol>
  *
  * <p>
- *   Static properties should be cached as a field for fast access.
- *   Dynamic properties get be accessed {@link #getProperty(Key, Class, Object)}
+ *   Static properties are snapshotted as primitive fields for fast access. Dynamic fields
+ *   are exposed as {@link Property} accessors so callers see live FP updates on every
+ *   {@code .get()}.
  *
  */
 public final class EVCacheTranscoderProperties {
@@ -43,9 +42,10 @@ public final class EVCacheTranscoderProperties {
     }
 
     private final String appName;
-    private final PropertyRepository propertyRepository;
 
     private final boolean binarySerializationEnabled;
+    private final int maxDataSizeBytes;
+    private final int compressionThresholdBytes;
 
     /**
      * Construct the bundle and snapshot every property via the three-level resolution chain.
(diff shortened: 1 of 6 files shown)


joegoogle123:
@janewang1680 once this merges you can rebase off of this for your compression PR.

evcache-core/src/main/java/com/netflix/evcache/config/EVCacheTranscoderProperties.java
```diff
@@ -18,8 +16,9 @@
  * </ol>
  *
  * <p>
- *   Static properties should be cached as a field for fast access.
- *   Dynamic properties get be accessed {@link #getProperty(Key, Class, Object)}
+ *   Static properties are snapshotted as primitive fields for fast access. Dynamic fields
```
shy-1234, line 19: Not a blocker since it's the original behavior, but just curious how much overhead would it be to also make those static properties dynamic? (I think the dynamic values are also cached?)

shy-1234 (approved):

evcache-core/src/main/java/com/netflix/evcache/config/EVCacheTranscoderProperties.java
joegoogle123: There are two fields that where static e.g  `max` and `compressionThreshold` which I don't think we can make dynamic. 

There is a new field to `binarySerializationEnabled` which we __could__ make dynamic but I did not measure the overhead of dynamically checking this property.

